PR #36, opened 2014-11-07
Adding independent values for R, G and B brightness.

gpambrozio:
When doing a project that uses 3V batteries I noticed that the blue gets very weak when voltage drops. Having independent values of brightness for R, G and B and making them different (in my case 20, 32 and 128) solved the problem.

This is backwards compatible as it still has the setBrightness method that takes one parameter. The only problem I solved in a non ideal way was the getBrightness method.

### DIFF
--- a/Adafruit_NeoPixel.h
+++ b/Adafruit_NeoPixel.h
@@ -56,6 +56,7 @@ class Adafruit_NeoPixel {
     setPixelColor(uint16_t n, uint8_t r, uint8_t g, uint8_t b),
     setPixelColor(uint16_t n, uint32_t c),
     setBrightness(uint8_t),
+    setBrightness(uint8_t r, uint8_t g, uint8_t b),
     clear();
   uint8_t
    *getPixels(void) const,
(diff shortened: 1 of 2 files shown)


Adafruit_NeoPixel.h
```diff
@@ -56,6 +56,7 @@ class Adafruit_NeoPixel {
     setPixelColor(uint16_t n, uint8_t r, uint8_t g, uint8_t b),
     setPixelColor(uint16_t n, uint32_t c),
     setBrightness(uint8_t),
+    setBrightness(uint8_t r, uint8_t g, uint8_t b),
     clear();
   uint8_t
    *getPixels(void) const,
```
hiren-jogani: in this function const means??

gpambrozio: From http://stackoverflow.com/questions/2157458/using-const-in-classs-functions:

"Typically you want to declare all non-mutating non-static class methods as const. This allows calling code to use the const qualifier on pointers, and it helps catch mistakes."

PR #42, opened 2015-02-26
Support for nRF51822 based boards

sandeepmistry:
Add support for the following [Nordic Semiconductor nRF51822](https://www.nordicsemi.com/eng/Products/Bluetooth-Smart-Bluetooth-low-energy/nRF51822) based boards:
 * [RedBearLab](https://www.nordicsemi.com/eng/Products/Bluetooth-Smart-Bluetooth-low-energy/nRF51822)
  * [BLE Nano](http://redbearlab.com/blenano/)
  * [nRF51822](http://redbearlab.com/redbearlab-nrf51822/)
 * [RFduino](http://www.rfduino.com)

Base on tolson's RFduino forum post:  http://forum.rfduino.com/index.php?topic=787.0

Tested with [NeoPixel Ring - 16](https://www.adafruit.com/products/1463) at NEO_KHZ800 and NEO_KHZ400, on RBL nRF51822 and RFduino.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -56,7 +56,7 @@ Adafruit_NeoPixel::Adafruit_NeoPixel(uint16_t n, uint8_t p, uint8_t t) : numLEDs
     gOffset = 1;
     bOffset = 2;
   }
-  
+
 }
 
 Adafruit_NeoPixel::~Adafruit_NeoPixel() {
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -823,6 +823,77 @@ void Adafruit_NeoPixel::show(void) {
 #error "Sorry, only 48 MHz is supported, please set Tools > CPU Speed to 48 MHz"
 #endif
 
+#elif defined(NRF51) || defined(__RFduino__)
+
+#ifdef NRF51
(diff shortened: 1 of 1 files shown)


microbuilder:
Hi Sandeep ... I unfortunately don't have access to any of the HW listed here, but are you running this with the radio enabled and some sort of connection or advertising going on over BLE?  I put quite a bit of effort into try to get reliable nRF51+NeoPixel timing with the radio enabled but even with careful assembly code wasn't able to respect the timing because the SoftDevice has higher priority and always causes gaps in the GPIO output once it starts sending an advertising frame, etc. Basically, anything real time is impossible when the radio is enabled (by design on Nordic's part).

This would of course be no problem when the radio isn't running, but that's not a terribly useful situation for an SoC designed solely for BLE.  I was just hoping for some confirmation that this worked for you reliably with the radio enabled and running.  I ended up using DotStar LEDs myself in combination with the radio since the timing there is compatible with the inevitable gaps in the IO.

sandeepmistry:
@microbuilder yes, I had the radio on and had a characteristic to set the RGB value of all LED's.

I was using the following sketch: https://gist.github.com/sandeepmistry/d2e8687a6993e9a110b7

BLEPeripheral doesn't set up the nRF51822 in interrupt mode, it polls instead. The patch also disables interrupts prior to sending data to the NeoPixel and then re-enables then after.

What pixel size were you testing with?

PR #46, opened 2015-04-01
Fix harmless compiler warnings

PaulStoffregen:
These compiler warnings are harmless, except for the psychological effect they have on customers when something else about their project doesn't work.


ladyada:
thx!

PR #58, opened 2015-07-27
ESP8266 fix and addition of setPixels();

forkineye:
`esp8266.c` requires inclusion of `eagle_soc.h` to compile in the latest ESP8266 Arduino environment.
`setPixels(uint8_t *buffer)` is used to quickly fill a pixel object from a contiguous RGB buffer (no color order mapping).
A dimming map (pseudo gamma correction) is also provided in `esp8266.c` via an optional #define, based off findings from this page - http://ohararp.com/tcs3471-color-sensor-and-ws281112-led-output/


toddtreece:
@forkineye thanks for the contribution! could you remove the `.esp8266.test.skip` files from the example  sketch folders on your fork to make sure the travis ci builds pass with the ESP8266?

forkineye:
@toddtreece I checked my fork and don't see the files which you reference, however did notice that travis skipped them.  Not sure what's going on with that.

toddtreece:
@forkineye it looks like you would need to pull in the latest changes from upstream, and then remove the files and commit after that for travis to verify esp8266.

thanks!

forkineye:
Who snuck those in there?! :)  Fixed and Travis is doing its thing now.

PR #60, opened 2015-08-08
Two changes to the Adafruit_NeoPixel::setPixelColor methods

t3chguy:
+ Allow for negative `n` to target LEDs from the end of the strip
+ Allow for overflow so:
    + lazier loops can be written
    + code can easier be more friendly with different length strips
    + better than silently not doing anything if `n` is out of the range as it did previously

All comments welcome, this will not impact any existing users if they are using the lib remotely correctly but will add functionality to the existing methods.


t3chguy:
Whoops spotted a tiny logic error, gonna fix that straight away

t3chguy:
I am wondering whether to also add this behaviour to `Adafruit_NeoPixel::getPixelColor`

PR #63, opened 2015-10-07
setBrightness(0) error

funkfinger:
I'm seeing what I believe to be a bug in setting brightness. The following sketch illustrates what I'm seeing - [gist](https://gist.github.com/440aeec8a6fe38d8e69f.git)

    // example sketch to demonstrate percieved neopixel library error.
    // expect green > off(-ish) > blue > off > red > off > green > off(-ish) > blue > off > red
    // get    green > off(-ish) > blue > off > red > off > off   > off       > blue > off > red

    #define PIN 0
    Adafruit_NeoPixel strip = Adafruit_NeoPixel(1, PIN, NEO_RGB + NEO_KHZ800);

    void setup() {
      pinMode(PIN, OUTPUT);
      strip.begin();
      strip.setPixelColor(0, strip.Color(255,0,0));
      strip.show();
    }

    void loop() {
      uint16_t wait = 2000;
      strip.setPixelColor(0, strip.Color(0,255,0)); // green
      strip.setBrightness(255);
      strip.show();
      delay(wait);
      // setting brightness to 1 is OK but pixel not 100% off
      strip.setBrightness(1);
      strip.show();
      delay(wait);

      strip.setPixelColor(0, strip.Color(0,0,255)); // blue
      strip.setBrightness(255);
      strip.show();
      delay(wait);
      // setting color to 'black' is ok...
      strip.setPixelColor(0, strip.Color(0,0,0));
      strip.show();
      delay(wait);

      strip.setPixelColor(0, strip.Color(255,0,0)); // red
      strip.setBrightness(255);
      strip.show();
      delay(wait);
      // but setting brightness to 0 kills strip
      strip.setBrightness(0);
      strip.show();
      delay(wait);
      // and we never get back brightness until we set pixel color
    }

A simple ternary in Adafruit_NeoPixel::setBrightness(uint8_t b) seems to fix the issue, but I'm not aware of other edge cases that it may affect.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1187,6 +1187,7 @@ void Adafruit_NeoPixel::setBrightness(uint8_t b) {
       *ptr++ = (c * scale) >> 8;
     }
     brightness = newBrightness;
+    // brightness = newBrightness == 1 ? 0 : newBrightness;
   }
 }
 
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1186,8 +1186,7 @@ void Adafruit_NeoPixel::setBrightness(uint8_t b) {
       c      = *ptr;
       *ptr++ = (c * scale) >> 8;
     }
-    brightness = newBrightness;
-    // brightness = newBrightness == 1 ? 0 : newBrightness;
+    brightness = newBrightness == 1 ? 0 : newBrightness;
   }
 }
 


PaintYourDragon:
Behavior is normal. Avoid multiple setBrightness() calls; it's meant as a one-shot. From the documentation here: https://learn.adafruit.com/adafruit-neopixel-uberguide/arduino-library

> setBrightness() was intended to be called once, in setup(), to limit the current/brightness of the LEDs throughout the life of the sketch. It is not intended as an animation effect itself! The operation of this function is “lossy” — it modifies the current pixel data in RAM, not in the show() call — in order to meet NeoPixels’ strict timing requirements. Certain animation effects are better served by leaving the brightness setting alone, modulating pixel brightness in your own sketch logic and redrawing the full strip with setPixel().

funkfinger:
Is there any interest in taking a stab at refactoring to accommodate a "more than once" setting of the global brightness value? Also, I was wrongly using the "source code as documentation" which should probably also reflect the above statement from the docs.

PaintYourDragon:
It doesn't much seem worth the effort, as it would require a complete rewrite of all the AVR assembly code into something even more incoherent.

PR #66, opened 2015-10-26
Add Implementation for defined function setPixelColor(n,r,g,b,w)

rtepowers:
## Overview

Noticed that the header has a line for 'setPixelColor(uint16_t n, uint8_t r, uint8_t g, uint8_t b, uint8_t w)' but there was no implementation.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1148,6 +1148,14 @@ void Adafruit_NeoPixel::setPixelColor(
   }
 }
 
+// Set pixel color and brightness from RGB and Brightness components
+void Adafruit_NeoPixel::setPixelColor(uint16_t n, uint8_t r, uint8_t g, uint8_t b, uint8_t w) {
+  if(n < numLEDs) {
+    this -> setBrightness(w);
+    this -> setPixelColor(n, r, g, b);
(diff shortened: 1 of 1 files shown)


mscosti:
I noticed that the other day too, glad to see a fix for it. 

Looks good to me! :+1:

PaintYourDragon:
Added this in most recent commit. Implementation was off; 'w' is 4th value for WRGB NeoPixels, not brightness.

rtepowers:
This is a little nitpick but wouldn't an API of setPixelColor(n, w, r, g, b) make more sense for WRGB pixels? I appreciate the clarification of this function either way.

ladyada:
we call em RGBW :)

PaintYourDragon:
The order is arbitrary and doesn't necessarily even correspond to what's going out on the wire. I just tacked it on at the end. Sometimes they're called WRGB pixels, sometimes RGBW, neither really has any bearing on what's going on inside, except there's a fourth element somewhere.

Is the zebra white with black stripes, or black with white stripes?

ladyada:
> Is the zebra white with black stripes, or black with white stripes?

/me has achieved enlightenment

rtepowers:
Awesome.

PR #67, opened 2015-10-29
Set Default Brightness Level

rtepowers:
The NeoPixels are incredibly bright. This change tones down the brightness for this particular example. As an added bonus, this commit also utilizes the 'setBrightness' method which helps demonstrate its functionality.

### DIFF
--- a/examples/strandtest/strandtest.ino
+++ b/examples/strandtest/strandtest.ino
@@ -28,6 +28,7 @@ void setup() {
 
 
   strip.begin();
+  strip.setBrightness(50); // Ranges from 0 (off) to 255 (full power)
   strip.show(); // Initialize all pixels to 'off'
 }
 


mscosti:
I was also thinking it could be interesting if we had a new animation function, possibly named 'pulse'.  It would basically animate through the whole brightness range from 0 - 255 (or whatever looks best), for all red, then all green, then all blue.

Besides looking cool it might bring more attention to setBrightness when people see it actively being changed when they run the example the first time.

Definitely keep it defaulted to lower than 255 like you have already though.

Thoughts?

UIOP456y:
Thank you for the brightness changes for the "strandtest".  Works great,
give me a little insight into the coding, still have alot to learn.

Thanks again

tablatronix:
:+1:

ladyada:
@tablatronix  plz dont +1 issues, it fills my inbox and doesn't make me code faster ;) please use the "reactions" button on the issue if you want to 'vote' on something

tablatronix:
Word.
Reactions are not available on mobile web version, I forget about them. I think you can leave watch on but disable emails... somewhere.

JohannesWilde:
Can this be closed? setBrightness() is now part of the strandtest example.

PR #69, opened 2015-11-01
use static calls in  Wheel

slootsky:
modify Wheel to use static calls to Colour, enabling the code to be
reused.


t3chguy:
Maybe it could even be shunted into the Lib itself rather than just an example function. It is a handy Method to have in my opinion anyway.

slootsky:
Not sure if that's required.  This also has the bonus of showing people how to call a static method when there isn't an object.

tablatronix:
:+1:

PaintYourDragon:
Making a sweep through old PRs. Always grateful for contributions and the exchange of ideas, but won’t be merging this, Wheel() is now deprecated in favor of ColorHSV(). Additionally, while the use of static calls is informative, that’s more a job for a programming tutorial and would distract from the example’s purpose of simply demonstrating lib functionality. Thank you.

PR #70, opened 2015-11-03
Added function for getting the pixel color for a single channel.

amicoleo:
New method that adds function for getting pixel color for a specific channel. 

I added an example showing possible use of it. I don't have RGBW neopixels laying around, so I haven't tested with them.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1224,6 +1224,53 @@ uint32_t Adafruit_NeoPixel::getPixelColor(uint16_t n) const {
   }
 }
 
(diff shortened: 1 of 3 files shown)


tablatronix:
Can you not just do a getpixelcolor return channel << 16,8 etc.

amicoleo:
Sure. It's just a helper function that makes it easier for those not too at ease with bitshift operators...

PaintYourDragon:
Making a sweep through some old PRs today. Grateful for the contribution, but won’t be merging this — equivalent functionality can be done in user code with a shift and mask operation. Thank you.

PR #71, opened 2015-11-14
expose the pin the Adafruit_NeoPixel is driving. (pin cannot be changed)

patmoore:
I needed this because I am passing the Adafruit_NeoPixel structure to functions that use the pin as an index into an array.


patmoore:
This pull request has some warnings that are preexisting. It does fix one existing warning in setPin

patmoore:
Note: getPin() returns by value not by reference, so pin cannot be changed.

ladyada:
added in a different pr!

PR #72, opened 2015-11-14
Create a typedef neoPixelColor to allow type clarity.

patmoore:
I created the neoPixelColor to keep the color values separated from other uint32_t. This helped straighten out a few bugs in my code.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1171,7 +1171,7 @@ void Adafruit_NeoPixel::setPixelColor(
 }
 
 // Set pixel color from 'packed' 32-bit RGB color:
-void Adafruit_NeoPixel::setPixelColor(uint16_t n, uint32_t c) {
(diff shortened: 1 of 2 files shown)


patmoore:
The warnings are preexisting and are fixed in other pull requests

tablatronix:
:+1:

PaintYourDragon:
Howdy. Always grateful for contributions and the exchange of ideas, thank you. Reflecting on this one a bit, I’ve decided against merging. If the functionality is beneficial in your own code, I’d recommend adding the typedef in your sketches, or you’re always welcome to create a fork which can easily be kept in sync with any other library changes.

Mostly it has to do with our aim to keep sketch code largely copy-and-paste interchangeable between NeoPixels and Dotstars, aside from the constructor. Any code that leans on this typedef won’t work with the other library, unless both headers are included. So, while maybe uint32_t isn’t optimal, it’s not exactly pessimal either, maybe just one of those awkward design corners that we’ve painted ourselves into.

PR #73, opened 2015-11-14
fix travis compiler warnings

patmoore:
I believe this will fix compiler warnings that travis is reporting on the existing code.


patmoore:
I have a better version coming

PR #82, opened 2015-12-21
added function to set only one color channel without touching the others

gregersn:
Added the possibility to easily change the value of one channel on a pixel, instead of having to set all three.


tablatronix:
Might as well add white channel also

PR #90, opened 2016-02-08
Fix compiler warnings

PaulStoffregen:
This fixes several harmless warnings, reported here:

https://forums.adafruit.com/viewtopic.php?f=47&t=89219&p=450053#p449325

I tested this code on Arduino Uno and Teensy-LC, even though these edits don't actually change the code.  Both worked fine.


ladyada:
thx!

PR #92, opened 2016-02-23
Arduino/Genuino 101 Support

bbaltz505:
Rebasing Arduino/Genuino 101 support changes onto the latest master.


bbaltz505:
Amended commit with compile warning fix

PR #94, opened 2016-03-15
Added getPin accessor method

brad95411:
How was this not already a thing? If you have the capacity to change the data you should be able to read what is already there. It would be helpful to have in my current projects. Which references data and functions for certain strips based on the data pin that they are connected to. It would be easier than trying to keep track of each strip's pin using an array, and require no excess memory overhead other than what's required to store this function.

### DIFF
--- a/Adafruit_NeoPixel.h
+++ b/Adafruit_NeoPixel.h
@@ -135,7 +135,8 @@ class Adafruit_NeoPixel {
     updateType(neoPixelType t);
   uint8_t
    *getPixels(void) const,
-    getBrightness(void) const;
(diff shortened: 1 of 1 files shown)


PaintYourDragon:
Good point. Doesn't compile as-is though. Add semicolon after 'return pin' and I'll merge this. Thanks!

brad95411:
Not sure why the check is failing this time, compiles fine in an Arduino environment for both Leonardo and Teensy 3.1/3.2 devices.

ladyada:
was added earlier

PR #114, opened 2016-11-12
Add HSV to RGB conversion to set pixel colors and an example

BsAtHome:
Implement (integer) HSV to RGB conversion with a very fast algorithm (see http://www.vagrearg.org/content/hsvrgb). Using the HSV color-space is significantly easier than the RGB color-space when gradients need to be created (f.ex. rainbows).

Added one function in the class: setPixelColorHsv() and the corresponding implementation. Defines are added to delimit the parameters' ranges. The implementation is cross-platform and should be functional for any usual platform and architecture (8-bit, 16-bit, 32-bit and 64-bit). An example is included to show its use.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1608,6 +1608,104 @@ void Adafruit_NeoPixel::setPixelColor(uint16_t n, uint32_t c) {
   }
 }
 
+// Set pixel color from HSV colorspace:
+// See http://www.vagrearg.org/content/hsvrgb for in depth algorithm and
+// implementation explanation.
+void Adafruit_NeoPixel::setPixelColorHsv(
+ uint16_t n, uint16_t h, uint8_t s, uint8_t v) {
+  if(n >= numLEDs)
+    return;
+
+  uint8_t r, g, b;
+
+  if(!s) {
+    // Monochromatic, all components are V
+    r = g = b = v;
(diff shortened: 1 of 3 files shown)


patmoore:
I would love to see this functionality -- it seems like this patch has lots of problems though (from the travis output)

BsAtHome:
I'd like to see it included too. But, well, the errors are none of mine...

The errors relate(d) to a name-collision of the "gamma" variable in the RGBWstrandtest example. I've renamed the array and made it the "correct" type. Additionally, I disabled the test for RGBWstrandtest on the trinket because it fails due to another name-collision and some more problems.

Now the build should be clean on all other platforms, including the added hsvrainbow example (builds clean on all platforms).

Hope these fixes help the cause ;-)

BsAtHome:
Is any of the maintainers actually listening in on the issues and PRs?

It is awfully quiet in here!

HardProg:
hello, your addendum helped me a lot, and I want to improve it a bit, namely: to make syntax highlighting setPixelColorHsv 
[keywords.txt](https://github.com/adafruit/Adafruit_NeoPixel/files/2297238/keywords.txt)

PaintYourDragon:
Making a sweep through some old PRs today. Grateful for the contribution, but a different PR somewhere along the way implemented ColorHSV() for similar functionality. Thank you.

PR #119, opened 2016-12-07
Make example compatible with Trinket

jpralves:
Removed Serial initialization - not needed and not working with Trinket
Changed array type from "int" to "byte" to fit in Atttiny85's Trinket.

### DIFF
--- a/examples/RGBWstrandtest/RGBWstrandtest.ino
+++ b/examples/RGBWstrandtest/RGBWstrandtest.ino
@@ -11,7 +11,7 @@
 
 Adafruit_NeoPixel strip = Adafruit_NeoPixel(NUM_LEDS, PIN, NEO_GRBW + NEO_KHZ800);
 
-int gamma[] = {
+byte gamma[] = {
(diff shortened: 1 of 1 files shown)


BsAtHome:
Travis CI fails still fails on this change. See my comment in PR #114 on this.

The first problem is with the variable name "gamma", which conflicts elsewhere. Changing the type to 'byte' does not change this. I changed the name to "mygamma" in PR #114 to get around the problem.

The second problem is that the define of "BIN" conflicts for the trinket compile because the stream library header is included by default in Arduino-type builds, which, via Print.h already defines a "BIN" for printing. Removing the Serial initialization does not change this for the CI build.

The specific builds on the user's computer seems to differ from Travis builds, which may be related to how Travis is set up, or updates elsewhere. Changing the way Travis does the builds is something the upstream maintainers may need to resolve.

BTW, Travis CI builds broke somewhere in November 2015, when looking at the PRs. This may be due to changes elsewhere, and not directly caused by additions in this specific repo. However, IMO, using CI and not keeping it up to date is worse than not having it at all.

PR #123, opened 2017-01-26
Neopixel silently does nothing on non supported architectures

marcmerlin:
After troubleshooting on an ESP32 and wondering why it wasn't working, I realized that due to nested ifdefs (hard to see initially), no code was actually being run for that processor.
Since there is no C fallback that works everywhere due to timing issues, add a compile error to save time to the end user.


marcmerlin:
Looking at the build test error, those errors seem totally unrelated to my change although my change indeed is designed to make the build fail on non supported platforms.

ladyada:
heya this was already merged with the esp32 code - thx!

PR #129, opened 2017-02-17
Make `getPin` and `canShow` const methods

donkopotamus:
These methods are not expected to mutate state, so should be `const`


PaintYourDragon:
Making a sweep through some old PRs today. Appreciate the contribution but marking it closed. Looks like some other PR along the way covered the getPin() case. In another, canShow() changed where it might mutate state now. Thank you.

Issue #118, opened 2016-12-05
Travis CI fails on master


BsAtHome:
The master branch fails the Travis CI build test. There are several issues regarding the RGBWstrandtest example. There are name-collisions that prevent the build to succeed.

PR #114 (HSV addition) now includes the fixes needed to make Travis CI tests to succeed again.

Even is the HSV addition is rejected, then the Travis fixes should be applied to make continuous integration tests useful again.

--
Greetings Bertho

BsAtHome:
BTW, this is also related to issue #100.

BsAtHome:
The latest merge from PR #119 fixed one part of the Travis CI problem. The one left is that the RGBWstrandtest example will not build on the trinket. Disabling that specific test, by creating the file "examples/RGBWstrandtest/.trinket.test.skip" will make Travis happy.

The problem with the trinket test is a #define of "BIN" name-collision between the avr library (avr/include/avr/iotnx5.h) and the arduino library (hardware/arduino/avr/cores/arduino/Print.h; cascaded from Stream.h). This one may be rather hard to fix in a good manner if the Stream code cannot be excluded from being required. Therefore, disabling the test may be the simplest option for now.

ladyada:
ok!

BsAtHome:
I don't quite understand the reply "ok!" and subsequently closing the bug while Travis still fails. Is the fix rejected or not acceptable? Or, the fix is acceptable, but was not yet committed (maybe by mistake)?

ladyada:
hiya we commited but didnt push til today - thankx!

PR #134, opened 2017-03-22
Rgbwstrandfix

BsAtHome:
- Fix signed/unsigned comparison warnings in the RGBWStrandTest example.
- Do not use floating point. Calculations can be performed using integer calculations
- Do not run Travis CI on the (attiny85) trinket.

Removal of the floating point use makes the code compatible with the attiny85 trinket by reducing the code-size (from 103% to 84%). However, it cannot be compiled in one pass. The Arduino IDE can compile the example successfully, but the sketch must be compiled twice.

The reason for the compile problem is a conflict due to a name-collision of a #define between the Arduino libraries and the standard avr include files. The define of "BIN" is conflicting in Arduino's Print.h header file, which cannot be solved easily (or at all) without altering the Arduino libraries (see also issue #118). The define is part of the serial print code, which is moot for the attiny85 based trinket.

The Ardiuno IDE ignores the errors on second compile (as long as you do not change the target board) because it has created a local cache-library of the core Arduino libraries. However, this does not work for Travis CI, because it will "see" the error on first compile and bail out. Therefore, the test must be disabled in the git repo, but the example is still functional.

### DIFF
--- a/examples/RGBWstrandtest/RGBWstrandtest.ino
+++ b/examples/RGBWstrandtest/RGBWstrandtest.ino
@@ -89,22 +89,22 @@ void pulseWhite(uint8_t wait) {
 
 
 void rainbowFade2White(uint8_t wait, int rainbowLoops, int whiteLoops) {
-  float fadeMax = 100.0;
-  int fadeVal = 0;
+  const unsigned fadeMax = 100;
+  unsigned fadeVal = 0;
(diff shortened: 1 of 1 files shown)


PaintYourDragon:
Making a sweep through some old PRs today. Closing as it appears that these changes, or equivalent, were incorporated somewhere along the way. Thank you!

PR #154, opened 2017-12-22
Repeat pixel array for nearly infinite LED strings with limited RAM

jerkey:
These changes allow addressing over 100,000 WS2812 LEDs from a normal ATMega328, by adding a "repeats" parameter to the Adafruit_NeoPixel initializer.  For example, if you have 1000 LEDs but only enough RAM for 100 LEDs, you can set a pattern of 100 LEDs and a "repeats" of 10, and it will control all 1000 LEDs.

These changes have been made only to the 16 MHz(ish) AVR implementation of ::show(void)

Adding the same functionality to the other implementations can be done in a style agreeable to the merging parties, everyone likes things done their own way :)

This code has been tested and seems to work properly (and is backwards compatible :)

I understand this PR is not 100% complete and my style may be imperfect, but i think a lot of users out there would benefit greatly by being able to control unlimited-length LED strings without using up any more RAM than the size of their desired repeating pattern.  Or has someone else already made this?

-jerkey

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -163,7 +163,8 @@ void Adafruit_NeoPixel::show(void) {
    *ptr = pixels,   // Pointer to next byte
     b   = *ptr++,   // Current byte value
     hi,             // PORT w/output bit set high
-    lo;             // PORT w/output bit set low
+    lo,             // PORT w/output bit set low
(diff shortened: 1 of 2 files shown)


ladyada:
hiya thanks for the PR - we can't have code that only works on one platform, so closing this PR! 
feel free to keep using your fork and thank you for the contribution! :)

PR #156, opened 2018-01-22
Support of BBC micro:bit

a-t-1:
Hi,

I couldn't find support in the Adafruit_NeoPixel library for the awesome BBC micro:bit. I did find the Microsoft PXT/neopixel (JavaScript?) sendBufferASM function that runs the neopixels on the make:code website. (https://github.com/Microsoft/pxt-neopixel/blob/master/sendbuffer.asm)

The attached .cpp has the sendBufferASM function ported to the arduino IDE asm syntax, inserted just before
#else // Other ARM architecture -- Presumed Arduino Due
starting with
#elif defined (VARIANT_BBC_MICROBIT)

The original sendBufferASM is commented out (for verification). Very ugly now. I'm happy to clean it up but I thought this might help someone with more advanced skills verify my tinkering.

Differences between the new code and sendBuffer

the assignment of the bitmask, following
https://github.com/sandeepmistry/arduino-nRF5/blob/dc53980c8bac27898fca90d8ecb268e11111edc1/variants/BBCmicrobit/variant.cpp

the assignment of the *reg (similar to the other ARM architecture), following
https://github.com/sandeepmistry/arduino-nRF5/blob/dc53980c8bac27898fca90d8ecb268e11111edc1/cores/nRF5/SDK/components/device/nrf51.h
and
http://www.iot-programmer.com/index.php/books/27-micro-bit-iot-in-c/chapters-micro-bit-iot-in-c/47-micro-bit-iot-in-c-fast-memory-mapped-gpio?showall=1

lsrs, adds and subs in sendBufferASM were not supported; changed to lsr, add and sub (with no detrimental consequenses)

The ported code was tested with example "strandtest" and a 150 rgb led string on pins 1 and 2 of the micro:bit.

a-t-1

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1721,6 +1721,95 @@ void Adafruit_NeoPixel::show(void) {
   }
 #endif
 
+
+
+#elif defined (_VARIANT_BBC_MICROBIT_)
+  uint8_t          *p   = pixels,
+                    pix, count, mask;
+  int32_t         num = numBytes;
+  unsigned int bitmask = ( 1 << g_ADigitalPinMap[pin] );
+// https://github.com/sandeepmistry/arduino-nRF5/blob/dc53980c8bac27898fca90d8ecb268e11111edc1/variants/BBCmicrobit/variant.cpp
+
+
+  volatile unsigned int *reg = (unsigned int *) (0x50000000UL + 0x508);
+
+// https://github.com/sandeepmistry/arduino-nRF5/blob/dc53980c8bac27898fca90d8ecb268e11111edc1/cores/nRF5/SDK/components/device/nrf51.h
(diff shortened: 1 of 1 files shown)


kd8bxp:
I'm having an issue getting neopixels to work on Pin 15 (P15). I have a DF Robot Maqueen Robot 
https://www.dfrobot.com/wiki/index.php/Micro:Maqueen_Robot_Car(V2.0)_SKU:_ROB0148
which is using P15 for the 4 neopixels that are on board.  The provided makecode works, so I know the neopixels work.  I also verified that I can use neopixels with the example provided here https://learn.adafruit.com/micro-bit-lesson-3-neopixels-with-micro-bit/software
with the neopixels attached to Pin 2 - but I've been unable to get them to work on the robot. I feel like I am missing something somewhere, but I'm unsure what that might be. Any ideas?

PR #164, opened 2018-05-28
Fix color (RGB => GRB)

NicolasBonduel:
It's `GRB` not `RGB`. I was confused for a while and thought my led wasn't working properly.


NicolasBonduel:
Maybe it should have a comment explaining that above the line in the example?

PR #166, opened 2018-05-29
Added user-friendly support for commonly used colors

patrick9958:
I've assembled a list of 140 preset colors so that instead of typing, for example, `theaterChase(strip.Color(20,34,16), delayval)`, users can type `theaterChase(TEAL, delayval)`. Both options for typing colors are available to users and the function will work either way. I've also put all 140 colors into the `keywords.txt` file and marked them as `KEYWORD2` so users know it will be recognized as a valid color by Arduino. I've already written code and tested it on two separate LED ring models.

I have changed the `Adafruit_Neopixel.h` file only by adding the `#include "hexadecimal_colors_to_strings.h"` at the top. I have added a new `hexadecimal_colors_to_strings.h` file in the directory as well. This file defines a list of 140 commonly used colors along with their hex values.

On top of that, I added a new example file called `colors.ino` in which I demonstrate how these colors can be used.

### DIFF
--- a/examples/colors/colors.ino
+++ b/examples/colors/colors.ino
@@ -3,6 +3,8 @@
   #include <avr/power.h>
 #endif
 
+using namespace NeoColors;    // used for Adafruit default color values like MAROON, TEAL, YELLOW
+
 // Which pin on the Arduino is connected to the NeoPixels?
 // On a Trinket or Gemma we suggest changing this to 1
 #define PIN            13
--- a/hexadecimal_colors_to_strings.h
+++ b/hexadecimal_colors_to_strings.h
@@ -1,140 +1,145 @@
-#define BLACK			0x000000
-#define NAVY			0x000080
-#define DARKBLUE		0x00008B
-#define MEDIUMBLUE		0x0000CD
-#define BLUE			0x0000FF
-#define DARKGREEN		0x006400
-#define GREEN			0x008000
-#define TEAL			0x008080
-#define DARKCYAN		0x008B8B
-#define DEEPSKYBLUE		0x00BFFF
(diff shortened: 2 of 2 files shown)


patrick9958:
First of all, sorry for taking so long to get back to you. Secondly, when you remove the include reference to the hexadecimal_colors_to_strings.h in both the .h and .cpp file, the example (colors.ino) does not compile because TEAL, YELLOW, and MAROON are now not declared. I could put the library include for the hexadecimal_colors_to_strings.h file at the top of the example but that would kind of defeat the purpose of integrating it within the Adafruit Neopixel library. Let me know how you want me to move forward.

lights0123:
I'd like to see these constants namespaced, either by prefixing them with `NEO_` or doing something like this:

    namespace NeoColors {
      const uint32_t BLACK = 0x000000
      const uint32_t NAVY = 0x000080
    }

Because otherwise any user sketches with any of these names won't work anymore. I've seen a couple of LCD graphics libraries that have these same constants, and those sketches won't work anymore without modifying either library (or adding hundreds of `#undef`s). This also allows sketches to just add `using namespace NeoColors;` if they aren't using anything that has a conflict to just use the short name.

patrick9958:
I have the `#include <stdint.h>` at the top of the hexadecimal_colors_to_strings.h file because Arduino was not recognizing uint32_t as a type but the example colors.ino file still won't compile because the hexadecimals.h file isn't included anywhere

ladyada:
please ping when this compiles and passes travis CI

abhinuvpitale:
You need to include "hexadecimal_to_string.h" header

PR #168, opened 2018-07-05
Prevent copy and assignment of Adafruit_NeoPixel class

Hello1024:
Typical code which copies the neopixel class will lead to use of free()'d memory, leading to crashes and odd behaviour.

Eg.

Adafruit_NeoPixel a;
a = Adafruit_NeoPixel(100);
a.begin();
a.show();

This appears to work, but the pixels buffer is freed on the assignment line, since the right hand side is a temporary, which then gets assigned to a, and then destroyed.  The destructor frees the pixels buffer, thich is shared by and still in use by a.

We could resolve this by making an explicit copy/assign constructor which copies the pixels buffer, but considering this library runs on embedded devices, the buffer is large, and copying the neopixels library probably isn't intentional, instead this PR simply denies copying by making the copy constructor private.

We do not use the C++11 syntax to allow this to run even on very old arduino versions.

### DIFF
--- a/Adafruit_NeoPixel.h
+++ b/Adafruit_NeoPixel.h
@@ -177,6 +177,12 @@ class Adafruit_NeoPixel {
   uint8_t
     pinMask;       // Output PORT bitmask
 #endif
+
+  private:
+   // Disallow copy and assign, since the default operators do not
+   // correctly handle our pixels buffer.
+   Adafruit_NeoPixel(Adafruit_NeoPixel const&);
+   Adafruit_NeoPixel& operator=(Adafruit_NeoPixel const&);
 };
 
 #endif // ADAFRUIT_NEOPIXEL_H


PaintYourDragon:
C++ is not a garbage-collected language like Python or Java. The example given will call the constructor and destructor twice (once on the declaration line, once on the assignment) and will free both the pixel buffer and the object itself, twice. If a NeoPixel object is a local or temporary instance, use the C++ 'new' operator instead:

>   Adafruit_NeoPixel *a;
>   a = new Adafruit_NeoPixel(100);
>   a->begin();
>   a->show();
>   delete a;

Hello1024:
The failing tests actually show that the examples actually have the exact same bug - they suffer a use-after free bug!

Considering that, a new approach will have to be taken to not break all the existing code which unintentionally uses the default copy/assign constructor.

We either need to refcount the pixels buffer, or to actually copy it as part of non default copy & assign operators.

PaintYourDragon:
Examples build without error prior to change. All examples are using a global Adafruit_NeoPixel object, have confirmed no use-after-free is occurring, constructor is invoked once, destructor never invoked. Use new & delete for NeoPixel objects that are limited in scope and there's no problem.

PaintYourDragon:
Other way to create one limited in scope is to not use an assignment, i.e.:
>   Adafruit_NeoPixel a(100, 6, NEO_GRB + NEO_KHZ800);
Object and pixel buffer are valid and not freed within the scope of the function. Destructor is automatically called when the function returns. Prior example using 'new' is OK if the object needs to be persistent, requires explicit 'delete'.

Joapfel:
Had the same problem as described by @Hello1024 . I was using the assignment operator which lead to unexpected behavior. Until a friend helped me out by looking at the Adafruit_Neopixel source code. Took me 1,5 days to debug this. Would be interesting to know why the proposed PR was not accepted?

PR #175, opened 2018-09-10
new example and updated keywords

Adminius:
- example with "new" operator added
- HEX representaion of NEO_* types
- added all NEO_* types to keywords.txt


Adminius:
this example because of: https://github.com/adafruit/Adafruit_NeoPixel_ZeroDMA/blob/master/Adafruit_NeoPixel_ZeroDMA.cpp#L15-L27

PR #187, opened 2019-01-04
nrf52 fix out of bound memory

hathach:
fix out of bound memory with nrf52 port

This causes hang in a particular scenario reported here
 https://forums.adafruit.com/viewtopic.php?f=57&t=145067&p=720463#p720463


RealAbsurdity:
Thanks! This fixed a disconnect issue with nrf52 and neopixels for me

PR #192, opened 2019-04-05
Remove superfluous checks for free() arguments

UniQP:
free() documentation: If ptr is NULL, no operation is performed.


PaintYourDragon:
Oh my god apparently that’s been true since like C99 and I just never got the memo. Thank you.

PR #193, opened 2019-04-11
Added kendryte k210 support

baorepo:
Modified library so that it refuses to build on an unknown architecture
instead of building and then doing nothing.

Added kendryte k210 support, which might not be optimal, but works for me.
Test with  Seeed's https://github.com/Seeed-Studio/ArduinoCore-k210/ and
Sipeed Maix Bit board


siddacious:
Hi @baorepo thanks for your work! We would like to hold off on adding support until the K210 is more widely available and can more easily be tested and supported.

For now I'm going to close this PR but I would suggest that you keep a fork with your changes so that you can make them available for those who are interested and you can merge in any changes from this repo as needed.

PR #194, opened 2019-05-08
correct formulas for Due 

Bob-the-Kuhn:
~This PR implements empirically derived timing values for the DUE (SAM3X8E) board.~

~If F_CPU is 84MHz then the empirically derived values are used.  If different then the calculated values are used.~

This takes care of a problem I was seeing with a DUE + RAMPS_FD_V1 + Fysetc MINIPANEL12864 rev 2.1 system.  The neopixel portion of the display was not working due to missing pulses and pulses of 62.5nS (the resolution of my logic analyzer).


Bob-the-Kuhn:
I'm running Marlin 2.0.x.

Bob-the-Kuhn:
Changed the approach of this PR.  Went from hard coded numbers to a different formula for creating the numbers.  The formula approach is easier to understand and maintain.

PR #196, opened 2019-07-19
Add Teensy 4 support

KurtE:
I brought over @PaullStoffregen's change to this library as shown in the posting of the Teeny 4 beta into the current release.

All of the changes are under one #elif
#elif defined(TEENSYDUINO) && (defined(__IMXRT1052__) || defined(__IMXRT1062__))

This is in the one very large and many #if #elif ... blocks in the method:
Adafruit_NeoPixel::show() 

So should not be able to effect any other  devices.  Did try rebuilding for T3.2 and LC.

Tested on one board with one Neopixel...


ladyada:
thanks!

PR #197, opened 2019-07-28
Change pin arguments to 16 bits from at for 32 bit uC compatibility 

mikeshub:
There really isn't much to say about this PR. I have setup a 3d printer using the LPC1768 microcontroller using Marlin. Marlin uses this code for the Neopixels. It doesn't work with the pin being an 8 bit value. So this changes them to 16 bit values.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -171,10 +171,10 @@ void Adafruit_NeoPixel::updateType(neoPixelType t) {
 #if defined(ESP8266)
 // ESP8266 show() is external to enforce ICACHE_RAM_ATTR execution
 extern "C" void ICACHE_RAM_ATTR espShow(
-  uint8_t pin, uint8_t *pixels, uint32_t numBytes, uint8_t type);
+  uint16_t pin, uint8_t *pixels, uint32_t numBytes, uint8_t type);
 #elif defined(ESP32)
 extern "C" void espShow(
-  uint8_t pin, uint8_t *pixels, uint32_t numBytes, uint8_t type);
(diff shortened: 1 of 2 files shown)


ladyada:
hiya thank u - please make adjustments so travis.ci passes :)

mikeshub:
That got it.

mikeshub:
Discussion over at Marlin on this:

https://github.com/MarlinFirmware/Marlin/issues/14754

PaintYourDragon:
If I were feeling more uptight I’d suggest a NeoPixel_pin_t typedef (with #ifdef(s) for different architectures) to minimize the cognitive load on something like an ATtiny. But it’s 2019 and c’mon, ATtiny, really? This is fine as-is and I will merge.

p3p:
Ah well guess my LPC176x support got in by the backdoor was trying to keep it separate until the LPC176x Arduino framework was a little more stable and Arduino compatible.

mikeshub:
Seems to me that Marlin is bridging the gap between Arduino and mBed (ARMduino?). Hopefully some of the Marlin effort will get picked up by Arduino for even more hardware support. Thanks for adding this in.

I'll be happy to add the pin type def. Do you think an #ifdef __AVR_ATtiny85__ would be the right preprocessor check for ATTiny or is there a better define?

Would it make sense to just go ahead and change the type to signed for the pins since in the class the pin is assigned to a signed variable? Or would it make more sense to do something like NeoPixel_pin_t and NeoPixel_upin_t for the signed vs unsigned?

ladyada:
no need to adapt this, the attiny will just have to cope with the extra bytes of ram used - thats ok

mikeshub:
You say that now, just you wait until this ruins someones gadget! :)

PR #209, opened 2019-10-24
Add support for Arduino Nano 33 BLE boards

manchoz:
Add support for Arduino boards based on [nRF52840](https://www.nordicsemi.com/Products/Low-power-short-range-wireless/nRF52840):

- [Arduino Nano 33 BLE](https://store.arduino.cc/arduino-nano-33-ble)
- [Arduino Nano 33 BLE Sense](https://store.arduino.cc/arduino-nano-33-ble-sense)

Tested with generic WS2812 strip on Arduino Nano 33 BLE and Arduino Nano 33 BLE Sense.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1507,7 +1507,11 @@ void Adafruit_NeoPixel::show(void) {
 //    pwm->INTEN |= (PWM_INTEN_SEQEND0_Enabled<<PWM_INTEN_SEQEND0_Pos);
 
     // PSEL must be configured before enabling PWM
+    #ifdef ARDUINO_ARCH_NRF52840
+    pwm->PSEL.OUT[0] = g_APinDescription[pin].name;
+    #else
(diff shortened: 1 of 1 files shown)


Adafruit_NeoPixel.cpp
```diff
@@ -1544,6 +1548,7 @@ void Adafruit_NeoPixel::show(void) {
   }// End of DMA implementation
   // ---------------------------------------------------------------------
   else{
+#ifdef digitalPinToBitMask
```
ladyada: this is a bad ifdef technique, please use ARDUINO_ARCH_NRF52840 or similar

ladyada:
hiya needs a few fixes

Adafruit_NeoPixel.cpp
manchoz: Fixed with ARDUINO_ARCH_NRF52840.

manchoz:
Done. Let me know.

PR #211, opened 2019-10-27
Corrected brightness usage

ArminJo:
Since real brightness is (brightness -1) we should use this in all formulas. Otherwise it will lead to unexpected results especially for low brightness values.

Please merge this request, my [NeoPatterns ](https://github.com/ArminJo/NeoPatterns) library relies on correct values.
Thanks
Armin

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1548,7 +1548,7 @@ void Adafruit_NeoPixel::show(void) {
   }// End of DMA implementation
   // ---------------------------------------------------------------------
   else{
-#ifndef ARDUINO_ARCH_NRF52840     
+#ifndef ARDUINO_ARCH_NRF52840
     // Fall back to DWT
     #ifdef ARDUINO_NRF52_ADAFRUIT
       // Bluefruit Feather 52 uses freeRTOS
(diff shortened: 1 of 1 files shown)


PaintYourDragon:
This is already factored into the brightness math. Storing brightness as (actual desired brightness + 1) is a trick to let us use a single-cycle >>8 rather than many-cycle /255 (even on contemporary architectures). This DOES produce small rounding errors in certain combinations, but it's A) small, B) not hooked up to a pacemaker or Mars lander, and C) seemed a worthwhile compromise for a significant performance payoff. I’ve always discouraged use of the brightness function anyway, since it’s linear and (as currently implemented) a lossy operation. It’s convenient for a quick-and-dirty scaling, fine for casual projects, anything more serious would want to use its own gamma table(s) anyway.

ArminJo:
Hi,
First I want to emphasize, that the patch does not change the fabulous single-cycle >>8 trick!

But in the old implementation, computing brightness of a starting value of 0x80 using the "gamma" factors 1,2,4,8,16,32,64,128 leads to results of 1,1,2,4,8,16,32,64 and not the expected 0,1,2,4,8,16,128 (it took me three hours to recognize this).
And as far as I can see, you tend to define your colors in examples using 128 and not 127 as color value (as I do). 
I can understand, that it is only a quick-and-dirty solution, not worth the effort.
But why not make it at least consistent for all the poor programmers to better understand your code, since the actual desired brightness is (brightness -1) and so why not use this value in the computation?
Regards
Armin

PR #213, opened 2019-11-05
Fix build warnings

henrygab:
**Sscope of change**

* Commit 141f6c8 changes multiple `#ifdef X` statements that are not otherwise erroneous to use the style `#if defined(X)`
* The fix at line 1527 (commit 2a6ebf5) is a clear example of why `#ifdef X` is not preferred … specifically because it can cause inadvertent errors, such as is fixed by this single-line commit.
* The fix at line 1418 (commit b8c45cb) is a trivial fix to a compiler warning.

**Known limitations**
There does not appear to be an automated build process for all the variations of code that could be `#define`'d.  Thus, manual code review (and automated build, if available) would be both appreciated and recommended, if only to verify no missed parenthesis, compiler complaints, or the like.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1524,7 +1524,7 @@ void Adafruit_NeoPixel::show(void) {
     // But we have to wait for the flag to be set.
     while(!pwm->EVENTS_SEQEND[0])
     {
-      #ifdef ARDUINO_NRF52_ADAFRUIT || ARDUINO_ARCH_NRF52840
+      #if defined(ARDUINO_NRF52_ADAFRUIT) || defined(ARDUINO_ARCH_NRF52840)
       yield();
       #endif
     }
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1415,7 +1415,7 @@ void Adafruit_NeoPixel::show(void) {
 #endif
   };
 
-  for(int device = 0; device < (sizeof(PWM)/sizeof(PWM[0])); device++) {
+  for(unsigned int device = 0; device < (sizeof(PWM)/sizeof(PWM[0])); device++) {
     if( (PWM[device]->ENABLE == 0)                            &&
         (PWM[device]->PSEL.OUT[0] & PWM_PSEL_OUT_CONNECT_Msk) &&
         (PWM[device]->PSEL.OUT[1] & PWM_PSEL_OUT_CONNECT_Msk) &&
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -45,7 +45,7 @@
 
 #include "Adafruit_NeoPixel.h"
 
-#ifdef TARGET_LPC1768
+#if defined(TARGET_LPC1768)
   #include <time.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


ladyada:
yep been a challenge to break the habit from 20 yrs :)
thank you for the great PRs!

ladyada:
FYI we're starting to add clang-format to our CI scripts (we'll be moving from travis to github actions soon) - if you have other C linting suggestions we'll take em!

henrygab:
One suggestion is fairly simple:

For any library code (or Arduino BSP) that is expected to be used by clients' compilers (e.g., headers, source code that they compile, etc.), enable all compiler warnings.  This not only finds issues (such as the `#ifdef` that may have affected calls to `yield()`), but also allows the clients to enable high warning levels without being inundated with warnings in code they didn't write.  (It also can enhance Adafruit's credibility when users first try to enable more warnings, and Adafruit's libraries are clean.)

For GCC, this could be as simple as compiling with `-Wall -Wextra -Werror`, while defining the minimum standard you want to conform to, such as:
* `-std=c11`
* `-std=c17`
* `-std=c++11`
* `-std=c++14`
* `-std=c++17`
Optionally, you could also enable `-Wpedantic -pedantic-errors`, to emit diagnostics required by the corresponding C/C++ version, and to discover additional **_compile-time undefined behavior_** -- although you may want to only enable that for TravisCI builds that are marked as `allow_fail `

Do you want me to PR a change to [Adafruit_nRF52_Arduino](https://github.com/adafruit/Adafruit_nRF52_Arduino), which enables additional build jobs that enable additional warnings?  (wall clock time for PRs will not change, and the additional builds will not delay or fail overall build status.)  I'm already using it myself, but would love to not have to keep merging changings...

ladyada:
yah i wish we could enable compiler warnings within the arduino CLI but i dont believe that's possible on its own....i think we can do that at at the BSP level and have it as an option like we do for speed/optimizations

yeah please submit any PRs to nrf repo - when thach has time he can review and merge!

henrygab:
Hi @ladyada,

### Enabling -Wall using Arduino CLI

FYI, while you are correct that you cannot use the CLI to enable `-Wpedantic` and `-pedantic-errors`,  you **_can_** enable `-Wall` using the Arduino CLI. See the [relevant lines](https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/tools/build_all.py#L64-L65) now in the [Adafruit_nRF52_Arduino](https://github.com/adafruit/Adafruit_nRF52_Arduino/) repository.

IMHO, enabling `-Wpedantic` in CI is probably going overboard ... especially as the Arduino UI doesn't enable a way to get there.  However, I do recommend the occasion compile with it... sometimes it can expose subtle errors.

### Static Analysis

As for other static-analysis tools, it really depends on your intent....

##### Microsoft solutions (C/C++/.NET)

(_This section is outdated ... my personal use for SAL annotations was over a decade ago..._)

I've used Microsoft's SAL annotations, which enable some truly amazing analysis (including for device drivers), but requires lots of annotation.  It also wouldn't be nearly as helpful in Arduino, as I'm not aware of any plumbing to enable PREFAST/PREFIX analysis outside of Visual Studio for C/C++ code.

The biggest problem with static analysis is the complexity needed to get really deep analysis.  The SAL annotations resolved this by enabling annotations at function boundaries, annotations on structures (both for validity and variable length), and of course strings and generic pointers.  This allowed the analysis to "presume" the correctness for annotations on all functions other than a single function being currently analyzed, making for a manageable problem.  .NET code had some really amazing work in Code Contracts, that had similar functionality, but was much easier to grok.  

##### What's out there now
Unfortunately, I've lost track of external efforts to bring such depth of analysis to other systems / compilers.  I encourage use of linters, to the extent that they find actionable errors, or warnings that an end-user would see.

I've never used clang's built-in static analysis.

Although I've never used it, I have previously heard positive things about Coverity.  I saw just now that that they have GitHub integration(!).  It might be worth checking out, especially if they would enable some of your depots (_although created and maintained by a for-profit company_) to be scanned for free.  (_can't hurt to ask, right?_)

Hope that helps!

PR #218, opened 2019-12-01
Fix Nano 33 BLE support

odensc:
Hi --

I was having issues where there would be no output to the LED strip on my Nano 33 BLE Sense board. After some debugging, I found out this line was the issue. `ENABLE` seems to always be equal to 1. Removing this line fixed it and the library seems to work fine - I haven't noticed any adverse effects.

cc/ @manchoz - since you authored the original PR to add support, perhaps you can shed some light on whether this is a proper fix?

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -1416,8 +1416,7 @@ void Adafruit_NeoPixel::show(void) {
   };
 
   for(unsigned int device = 0; device < (sizeof(PWM)/sizeof(PWM[0])); device++) {
-    if( (PWM[device]->ENABLE == 0)                            &&
-        (PWM[device]->PSEL.OUT[0] & PWM_PSEL_OUT_CONNECT_Msk) &&
+    if( (PWM[device]->PSEL.OUT[0] & PWM_PSEL_OUT_CONNECT_Msk) &&
         (PWM[device]->PSEL.OUT[1] & PWM_PSEL_OUT_CONNECT_Msk) &&
         (PWM[device]->PSEL.OUT[2] & PWM_PSEL_OUT_CONNECT_Msk) &&
         (PWM[device]->PSEL.OUT[3] & PWM_PSEL_OUT_CONNECT_Msk)


manchoz:
**TL;DR** Please, upgrade to latest Arduino Mbed Core 1.1.3.

@odensc On Arduino 33 BLE boards, the Mbed core will enable PWM by default so you need to disable it in order to get the NeoPixel library working.

The Arduino Mbed Core [PR 28](https://github.com/arduino/ArduinoCore-nRF528x-mbedos/commit/0977bd4cb6604187c38f5c707208abfea28c5156) fix this AND has been already [published](https://github.com/arduino/ArduinoCore-nRF528x-mbedos/blob/master/variants/ARDUINO_NANO33BLE/variant.cpp#L93) to Arduino Mbed Core 1.1.3.

odensc:
@manchoz Hmm, I guess it was because I started tinkering with this before 1.1.3 was out. Confirmed updating it fixes the issue. Thanks!

ladyada:
nice thanks @manchoz !

manchoz:
You welcome, @ladyada

lflima:
Hello guys, I am trying to get the adafuit neopixel library working on a nano 33 BLE sense and a WS2813B led strip with no luck. Leds will no turn on, although sometimes with the theaterchase effect some of them turn on randomly.
I have a logic lecel converter connected between the arduino and the led strip, boards and library updated to the last version, checked all comments from other guys in forums and no luck. Same code and same setup in an arduino mega works fine.
Any ideas?

Regards

dhalbert:
The WS2813B strips don't respond to the same timing that the WS2812 strips do. We've had one or two other reports of issues. We don't sell the WS2813 strips so we haven't yet tried to adapt the timings to work on both.

lflima:
Thank you for your comment.
But it works flawlessly on the MEGA 2560.
Could this be an issue only on the nordic based arduinos?
Any ideas on how to try adjusting the timmings?

dhalbert:
i am ordering a sample and will try to get it to work.

lflima:
Ok, thank you.
Meanwhile I will keep on searching and post any updaste

PR #228, opened 2020-03-03
Fixing compiler warning

VanKurt:
This fixes the following compiler warning:

> Fixing compiler warning 'variable 'pinMask' set but not used [-Wunused-but-set-variable]'

To do so, the pinmask variable and its assignment have been moved to the ESP8266 #ifdef block.


ladyada:
yay thanks!

PR #229, opened 2020-03-04
implement nrf show() non-blocking

hathach:
- if show() is called again, it will blockingly wait for previous show() to complete first (if not yet).
- also drop the DWT fallback implementation since it is rarely used. We didn't run into scenario where there is no PWM device.

TODO: may piggy back multipel neopixel strip (instance) into 1 hw PWM later.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -49,7 +49,7 @@
   #include <time.h>
 #endif
 
-#if defined(NRF52) || defined(NRF52_SERIES)
+#if defined(NRF52_SERIES)
 #include "nrf.h"
 
 // Interrupt is only disabled if there is no PWM device available
@@ -71,9 +71,16 @@
 */
 Adafruit_NeoPixel::Adafruit_NeoPixel(uint16_t n, uint16_t p, neoPixelType t) :
   begun(false), brightness(0), pixels(NULL), endTime(0) {
+
+#ifdef NRF52_SERIES
+  _pwm_dev = NULL;
+  _pixels_pattern = NULL;
+#endif
+
   updateType(t);
   updateLength(n);
   setPin(p);
+
 }
 
 /*!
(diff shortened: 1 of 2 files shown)


PaintYourDragon (changes requested):
There’s a likely issue here with the non-blocking transfer and the interaction of the 'endTime' element and the canShow() function.

endTime is normally set at the end of a blocking data transfer. It indicates the time at which the last bit was actually transmitted. show() then returns, and calling code can proceed with other duties while the 300 microsecond NeoPixel latch time elapses…on the next call to show(), the code blocks (if necessary) until the latch is complete.

As currently written with non-blocking DMA, endTime is being set right after the DMA starts, so it contains the time of the first bit out, rather than the last. This might work with early NeoPixels (which had a shorter latch time) and/or shorter runs, but I anticipate problems with long strips and/or WS2812B pixels.

Can think of a couple ways this might be fixed…

- After DMA starts, set endTime to micros() + (total bits * 5 / 4) (allowing the 32-bit value to roll over, it will still work). This is only an approximation but probably close enough.
- Have DMA transfer completion trigger an interrupt, which sets endTime to micros(). This is tricky in that each NeoPixel instance has its own endTime variable, but interrupts usually exist outside the class scope.
(Ignore earlier idea about padding end with zero-bytes, since these still generate a short '1' bit. I deleted that here.)

Swap-File:
I found that under certain condition, on mbed based nRF52840 boards using Adafruit_NeoPixel and ArduinoBLE at the same time could cause the ArduinoBLE stack to lock up occasionally with the error "[Info] Unhandled event".

Specifically, I had a central device writing 15 byte payloads to a peripheral at 20hz, while also writing to a string of 20 neopixels at 50hz on the peripheral.

The lockup would occur when BLE data came in during a LED show(), but not always.

It seemed like the lockup would happen if BLE data came in while show() was configuring the PWM & DMA transfer.

Adafruit's show() function for the nRF52840 architecture reconfigures the PWM and DMA transfer every time it is called, which assuming the led strip length and type isn't updated, isn't needed.

Here's my hacked up copy of the library that only configures the PWM and DMA transfer once, on begin().

It fixed all my lockup problems, and immediately returns after being called.

https://github.com/Swap-File/neopixel_nRF52840_dma/

Issue #231, opened 2020-05-15
Latch timing reference can cause infinite loop


Hampycalc:
I am encountering an infinite loop. On reviewing the code, I have discovered the following possibility.

It is possible for the latching timing reference variable `endTime` ([defined here](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.h#L350)) to be set ([here](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.cpp#L2210)) to the value of `micros()`, then value of `micros()` to overflow and `bool canShow(void) const { return (micros()-endTime) >= 300L; }` ([defined here](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.h#L234)) to always return `false`. This results in an (almost) infinite loop in the `show()` function at [this line](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.cpp#L205).

As a fix, the `canShow()` function should set `endTime` to `micros()` if `endTime > micros()`.

Hampycalc:
This has a chance to happen every 4294967295/(1000000*60*60) = 1.2 hours

PR #232, opened 2020-05-15
Latch timing reference can cause infinite loop

Hampycalc:
Addresses: https://github.com/adafruit/Adafruit_NeoPixel/issues/231

I am encountering an infinite loop. On reviewing the code, I have discovered the following possibility.

It is possible for the latching timing reference variable `endTime` ([defined here](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.h#L350)) to be set ([here](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.cpp#L2210)) to the value of `micros()`, then value of `micros()` to overflow and `bool canShow(void) const { return (micros()-endTime) >= 300L; }` ([defined here](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.h#L234)) to always return `false`. This results in an (almost) infinite loop in the `show()` function at [this line](https://github.com/adafruit/Adafruit_NeoPixel/blob/766f7fba6bb48081eaf861933a94a25524bd7be1/Adafruit_NeoPixel.cpp#L205).

As a fix, the `canShow()` function should set `endTime` to `micros()` if `endTime > micros()`.

### DIFF
--- a/Adafruit_NeoPixel.h
+++ b/Adafruit_NeoPixel.h
@@ -37,16 +37,16 @@
 #define ADAFRUIT_NEOPIXEL_H
 
 #ifdef ARDUINO
-#if (ARDUINO >= 100)
-#include <Arduino.h>
-#else
-#include <WProgram.h>
-#include <pins_arduino.h>
-#endif
+  #if (ARDUINO >= 100)
+  #include <Arduino.h>
+  #else
+  #include <WProgram.h>
+  #include <pins_arduino.h>
+  #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


ladyada:
thanks :)

PR #251, opened 2020-10-28
Update RGBWstrandtest.ino

bitDealer:
The defintion of BRIGHTNESS wasn't used in the setup-routine, therefore changing its value didn't change the brightness as expected


ladyada (approved):

ladyada:
thanx!

Issue #139, opened 2017-05-16
ESP32: timing error every 1ms corrupts LED pixel data


MustBeArt:
On the ESP32, an extra delay of about 5 microseconds is inserted into the signal output every millisecond. This is far outside the Neopixel timing tolerance of 150 ns, so it leads to incorrect values in some of the pixels.

The data for up to 33 RGB pixels can be transmitted in less than one millisecond (1.25 uS per bit X 24 bits per pixel X 33 pixels = 990 uS). In the common case where the show() method call immediately follows a delay() function call, the problem is not encountered unless the strand has at least 34 RGB pixels. However, if the code keeps time some other way (such as by calling millis() repeatedly) the extra delay may occur with random alignment to the signal. In that case, the signal may be corrupted with even a single pixel strand.

The extra delay may also be avoided by disabling interrupts during the call to the show() method. In the ESP32 Arduino environment, this cannot be done in the usual way, by calling noInterrupts() -- that function is defined to be nothing in Arduino.h. It can, however, be accomplished by invoking the macro portDISABLE_INTERRUPTS() defined by FreeRTOS in the file portmacro.h, since the Arduino code in the ESP32 is actually running on top of FreeRTOS. The macro portENABLE_INTERRUPTS() re-enables interrupts. Of course, if this method is used, any other time-critical operations will suffer interference, probably including millisecond timekeeping.

## Method to Reproduce

The best way is to observe the output signal on a logic analyzer. Use any sketch that talks to pixels, and set the number of pixels to at least 34. In each burst of activity on the signal line representing a single strand update, you should see at least one pulse (high or low) of between 5 and 6 uS, and those pulses should occur every millisecond thereafter (when the millisecond coincides with signal activity).

If no logic analyzer is available, you can see the problem with just a strand of pixels. If a strand of more than 33 pixels is available, simply hook it up and try to run any of the example sketches. Flaky pixels will be visible.

If no long strand of pixels is available, the following sketch will illustrate the problem with whatever length strand you do have. Because the strand is short, the extra pulse will only occur some fraction of the time, so you may have to watch for a while to see flaky pixels. (This is how I originally discovered the problem. I use this method of timekeeping when I need to process more than one time-critical device.)

```
#include <Adafruit_NeoPixel.h>

unsigned long tickTime = millis();

#define PIN 17
#define FRAME_INTERVAL 50

// Parameters set for the 24-Neopixel ring of RGB (no white) pixels
[...]

PR #253, opened 2020-10-30
Add RMT support to ESP32

### LINKED ISSUE #159: Wrong color on first pixel with ESP32
- Arduino board:  **HUZZAH32 (ESP32)**

- IDE version:  **platformio 3.5.2a7**

I have 2 pixels _(Breadboard-friendly RGB Smart NeoPixel)_ connected in series to the ESP32. No matter what I do, the first led _(pixel 0)_ is green, and if I set a color to it, it shows up different.

I switched the LEDs physically and still _pixel 0_ is green. FastLED, however, doesn't have this problem and shows the colors correctly.

I have tried with the latest code from this library, as well as with 1.1.3 and 1.1.4 and have the same problem.

```

Adafruit_NeoPixel pixels = Adafruit_NeoPixel(2, 21, NEO_GRB + NEO_KHZ800);

void setup() {
  pixels.begin();
  pixels.setPixelColor(0, pixels.Color(0, 0, 0));
  pixels.setPixelColor(1, pixels.Color(0, 0, 0));
  delay(10);
  pixels.show();
}

void loop() {}
```

hierophect:
Replaces the ESP32's espShow function with one that uses the RMT (Remote Control) peripheral to send timed pulses on the Neopixel line. Using the RMT should result in less flaky pixel performance when used with long Neopixel strings or in applications with heavy use of interrupts. Tested on a HUZZAH32 Featherwing.

This PR *should* resolve #159 and #139. However, I don't have a long enough string of Neopixels to reliably reproduce those errors on the original code, so any extra testing would be greatly appreciated to confirm this fixes all instability. This code should be easily portable to the ESP32-S2 and S3 when support is officially added to the Arduino core, since it is based off the existing implementation for the S2 in Circuitpython.

Note that the ESP8266 still uses the old timing method, as it does not have the RMT peripheral.

### DIFF
--- a/esp.c
+++ b/esp.c
@@ -155,6 +155,8 @@ void espShow(uint8_t pin, uint8_t *pixels, uint32_t numBytes, boolean is800KHz) 
     // Free channel again
     rmt_driver_uninstall(config.channel);
     rmt_reserved_channels[channel] = false;
+
+    gpio_set_direction(pin, GPIO_MODE_OUTPUT);
 }
 
 #endif
--- a/esp.c
+++ b/esp.c
@@ -1,0 +1,160 @@
+// Implements the RMT peripheral on Espressif SoCs
+// Copyright (c) 2020 Lucian Copeland for Adafruit Industries
+
+/* Uses code from Espressif RGB LED Strip demo and drivers
+ * Copyright 2015-2020 Espressif Systems (Shanghai) PTE LTD
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if defined(ESP32)
+
+#include <Arduino.h>
+#include "driver/rmt.h"
+
+// This code is adapted from the ESP-IDF v3.4 RMT "led_strip" example, altered
+// to work with the Arduino version of the ESP-IDF (3.2)
+
+#define WS2812_T0H_NS (400)
+#define WS2812_T0L_NS (850)
+#define WS2812_T1H_NS (800)
+#define WS2812_T1L_NS (450)
+
+#define WS2811_T0H_NS (500)
+#define WS2811_T0L_NS (2000)
+#define WS2811_T1H_NS (1200)
+#define WS2811_T1L_NS (1300)
+
+static uint32_t t0h_ticks = 0;
+static uint32_t t1h_ticks = 0;
+static uint32_t t0l_ticks = 0;
+static uint32_t t1l_ticks = 0;
+
+// Limit the number of RMT channels available for the Neopixels. Defaults to all
+// channels (8 on ESP32, 4 on ESP32-S2 and S3). Redefining this value will free
+// any channels with a higher number for other uses, such as IR send-and-recieve
+// libraries. Redefine as 1 to restrict Neopixels to only a single channel.
+#define ADAFRUIT_RMT_CHANNEL_MAX RMT_CHANNEL_MAX
+
+#define RMT_LL_HW_BASE  (&RMT)
+
+bool rmt_reserved_channels[ADAFRUIT_RMT_CHANNEL_MAX];
+
+static void IRAM_ATTR ws2812_rmt_adapter(const void *src, rmt_item32_t *dest, size_t src_size,
+        size_t wanted_num, size_t *translated_size, size_t *item_num)
+{
+    if (src == NULL || dest == NULL) {
+        *translated_size = 0;
+        *item_num = 0;
+        return;
+    }
+    const rmt_item32_t bit0 = {{{ t0h_ticks, 1, t0l_ticks, 0 }}}; //Logical 0
+    const rmt_item32_t bit1 = {{{ t1h_ticks, 1, t1l_ticks, 0 }}}; //Logical 1
+    size_t size = 0;
+    size_t num = 0;
+    uint8_t *psrc = (uint8_t *)src;
+    rmt_item32_t *pdest = dest;
+    while (size < src_size && num < wanted_num) {
+        for (int i = 0; i < 8; i++) {
+            // MSB first
+            if (*psrc & (1 << (7 - i))) {
+                pdest->val =  bit1.val;
+            } else {
+                pdest->val =  bit0.val;
+            }
+            num++;
+            pdest++;
+        }
+        size++;
+        psrc++;
+    }
+    *translated_size = size;
+    *item_num = num;
+}
+
+void espShow(uint8_t pin, uint8_t *pixels, uint32_t numBytes, boolean is800KHz) {
+    // Reserve channel
+    rmt_channel_t channel = ADAFRUIT_RMT_CHANNEL_MAX;
+    for (size_t i = 0; i < ADAFRUIT_RMT_CHANNEL_MAX; i++) {
+        if (!rmt_reserved_channels[i]) {
+            rmt_reserved_channels[i] = true;
+            channel = i;
+        }
+    }
+    if (channel == ADAFRUIT_RMT_CHANNEL_MAX) {
+        // Ran out of channels!
+        return;
+    }
+
+    // Match default TX config from ESP-IDF version 3.4
+    rmt_config_t config = {
+        .rmt_mode = RMT_MODE_TX,
+        .channel = channel,
+        .gpio_num = pin,
+        .clk_div = 2,
+        .mem_block_num = 1,
+        .tx_config = {
+            .carrier_freq_hz = 38000,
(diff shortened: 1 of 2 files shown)


ladyada:
neopixels look great!
however, if you have two neopixel objects (say, one internal pixel and some external strip) they all update the exact same, rather than being separate. so its some RMT allocation thingy. perhaps increment thru so folks have up to 10?

IAmOrion:
So do we just replace the current library with this version? As great as NeoPixelBus is, I found it much easier making animation classes (Extending on neopatterns) using the Adafruit Library so would be great to get this working without issues on ESP32's.  I will be controlling 2 independent strips (32 each strip) but am I understanding correctly from Ladyada's comment that this currently only supports 1 singular strip/object?  MTIA

ladyada:
yah if you can verify that your strip works (even if only one) we will fix it up to add multiple strip support when @hierophect is around next :)

IAmOrion:
Well, it certainly worked for me with a single strip yes :) I was able to use multiple NeoPatterns etc and control it from web page too.  I also re-uploaded sketch using Adafruit NeoPixel library to vreify that the RMT version was acting different and it most certainly was.  The Non-RMT version had some many artefacts or erroneous pixel behaviour it was annoying.  The RMT version worked great, albeit only for 1 strip

ladyada:
awesome, thak you for testing.

IAmOrion:
No problem.  I've been eagerly awaiting a fix as I prefer the Adafruit NeoPixel library over the NeoPixelBus library.  Mostly because it's what I'm used to , but I also prefer NOT having the animation api thast neopixelbus has, I prefer having the NeoPattern classes method etc so I look forward to this being accepted and an update soon :)

MustBeArt:
I'm excited that @hierophect has taken on this long-standing need. Thanks!

I'm also doing some testing, and I agree with most of what you've all found. It does solve the interrupt latency problem from my #139, and it seems to work fine with one strip, and it does not work correctly for multiple strips.

I get a slightly different result than @IAmOrion in the case of two strips. I find that activity intended for either strip ends up on both, for a test case where they are not trying to be active simultaneously.

The dynamic allocation of RMT channels in this code isn't actually doing anything. That's because it always blocks and waits for the transfer to complete, and then releases the channel allocation. So the next call can and does allocate that same channel again. (The channel it always allocates is 7, rather than the expected 0, because there's a ```break``` missing from the loop that looks for a free channel.)

That would be non-fatal if the new allocation and configuration completely overrode the old one, but evidently that's not the case. If we want to re-use an RMT channel, we need to figure out how to get it disconnected from the old pin. I have not gotten to the bottom of that rabbit hole, yet.

Blocking and waiting for transfer to complete is probably fine in a lot of cases, and is consistent with a lot of other Arduino libraries, but it breaks the promise made in a comment in ```Adafruit_NeoPixel::show()``` that _"multiple instances on different pins can be quickly issued in succession (each instance doesn't delay the next)"_ and hurts performance for multi-strip applications. I think we eventually need to avoid blocking. That's probably going to involve abandoning the default ISR and writing our own to handle the transmit done interrupt.

We could allocate RMT channels more statically, so that the number of available RMT channels limits the number of instantiated strips. That would avoid the need to re-use an RMT channel. Or, assuming we can figure out how to re-use an RMT channel, we can be more clever and allow unlimited strips (with blocking when all the RMT channels are busy). Neither option is trivial.

MustBeArt:
Turns out that, if my examination of the ESP IDF source code is correct, there is no way to share a RMT channel entirely through the RMT driver, because it has no way to remove the connection between the RMT channel and the pin.

You can, however, do it by going directly to the GPIO API, using the undocumented (?) function gpio_matrix_out() (which is in ROM!) and  its stealthy feature mentioned in gpio.h: if you pass in the magic value 0x100 for the signal index, it cancels output to the pin. This is what FastLED does in their RMT implementation.

So, a minimal repair for the problem with multiple strips is to add this line of code after the call to rmt_driver_uninstall:
```
gpio_matrix_out(pin, 0x100, 0, 0);
```
That doesn't achieve the performance possible by allowing multiple RMTs to run in parallel, but it does restore logical correctness for multiple strips, and it doesn't restrict the number of strips that can be instantiated. It works on my testbed, though I wouldn't call my testing thorough.

hierophect:
Was off for the weekend, I'll resolve the dual strip issue today.

ladyada:
welcome back & thanks :) if possible please support n=8 or more strips (if we can!) note that the RMT peripheral isnt really used for anything else so you can use it freely and we'll warn people that they can't also have other code with RMT

IAmOrion:
@hierophect Awesome, thank you so much for your work on this - I've been eagerly awaiting a fix for the Adafruit Official NeoPixel Library + ESP32 support using RMT.  Thanks again

hierophect:
Pushed a change that works for the ESP32S2 on Circuitpython, will test on ESP32 tomorrow once I get more Neopixel lines in. 

Since it is undocumented, I'm not sure why FastLED goes for `gpio_matrix_out(pin, 0x100, 0, 0);`, it may have simply been the best option in the API at the time they designed their implementation. I've found `gpio_set_direction` and `gpio_reset_pin` work just as well, though I'm waffling on which is more appropriate for this particular case. 

I understand the concern about performance. Circuitpython, which this implementation is based off of, tries to treat each Neopixel transaction as a discrete event that inits and de-inits itself in a single transaction. Since the expectation for Arduino is that it'll be a bit faster, I'll think about how to speed things up a bit in the absence of the pin and peripheral reservation system CPy uses.

IAmOrion:
This still seems buggy for me with 2 strips.  Some neopatterns don't function on 2nd strip at all, and passing a speed value does seem to make much difference.  Eg, I have a knight rider animation, but the speed difference between 200ms per led, and 15000 per led is minimal.  It's barely noticeable - but as you can imagine, I would've expected it to be going either really slow or super fast at those 2 different speeds

hierophect:
@IAmOrion thanks for testing! As I mentioned, I still don't have the hardware to test thoroughly on ESP32, so I'll take another look as soon as it comes in the mail today.

IAmOrion:
Excuse my sloppy code, it's a WIP but here's my project if you want to see what I was doing code/strip wise
[BMW_Z4_Wing_Mirrors_ESP32_AdafruitNeoPixel.zip](https://github.com/adafruit/Adafruit_NeoPixel/files/5482173/BMW_Z4_Wing_Mirrors_ESP32_AdafruitNeoPixel.zip)

MustBeArt:
@IAmOrion It looks like your speed parameters are squeezing through a uint8_t method parameter, so a value of 15000 isn't going to arrive intact. That probably explains your speed observations. Try values like 10 and 250 to see if you can see a big speed difference.

Can you construct a simplified example that just uses two Neopatterns on two strips and demonstrates the failure of some patterns to work on a second strip?

I was able to modify a version of the AllPatternsOnMultiDevices example sketch that ships with NeoPatterns and run it on two strips on ESP32, and everything appeared to work reasonably. Which patterns are failing for you?

IAmOrion:
@MustBeArt When I talk of NeoPatterns, I refer to my modified version of the NeoPatterns class file found in Adafruits "Multi Tasking Part 3" guide found here -> https://learn.adafruit.com/multi-tasking-the-arduino-part-3/put-it-all-together-dot-dot-dot

The knight rider speed issue will be my own stupidity there so will check that shortly.

I do have the NeoPatterns library that's not "Adafruit Official" as far as I know, so will give the examples a try and see how I get on with them.

hierophect:
I've tested this on an ESP32 feather with multiple strips of Neopixels, and things seem to be functioning as expected. Since @IAmOrion has confirmed they're no longer seeing issues, should be viable to merge, if the slight performance loss is deemed acceptable for now. I can investigate some performance fixes and follow up with another PR.

@ladyada, want to take another look?

ModsNSods:
The only animation I'm having trouble with in my code (See previous post for attachment) is that the POLICE pattern I made works fine on one strip but not on the other - the blue lights stay on permanently instead of flashing.  The animation is pretty fast and if I run the animation on 1 at a time, it works perfectly so I'm guessing that's potentially a good example of the slight performance issue.

That being said I plan to amend it anyways so will figure a workaround (or just re-code)

ladyada:
tested, this is good to go for now! can always improve later :)

jpasqua:
FYI: I have some WS2812D's that still have timing issues with this code (many other strips work fine). While testing and inspecting the code and comparing to the FastLED library, the ESP32-Digital-RGB-LED-Drivers, and the ESP-IDF code, it seems that this library does not send the final reset signal (low time).  Perhaps I'm missing it. In the other libraries this is achieved by pronging the duration1 value of the final rmt pulse pair to include the reset time (see examples below).

From ESP32-Digital-RGB-LED-Drivers:
```c
    // Handle the reset bit by stretching duration1 for the final bit in the stream
    if (i + pState->buf_pos == pState->buf_len - 1) {
      RMTMEM.chan[pStrand->rmtChannel].data32[i * 8 + offset + 7].duration1 =
        ledParams.TRS / (RMT_DURATION_NS * DIVIDER);
    }
```

From FastLED:
```c
        mBuffer[mCurPulse-1].duration1 = RMT_RESET_DURATION;
```

In both cases this is done in the `ws2812_rmt_adapter()` function (or equivalent). I noticed that the led_strip_rmt_ws2812.c code (which you mentioned that you adapted) is also missing the final reset, though it does have a constant defined for the reset time in microseconds: `WS2812_RESET_US`.

ladyada:
@jpasqua hmm! could you add a pull request to add it? we don't have any pixels that *dont* work :)

PR #266, opened 2021-04-11
Add RP2040 support using PIO

Bodmer:
This pull request adds support for the Raspberry PI Pico and RP2040 based processors. It uses the PIO for the serial data timing.

It has been tested with WS2812 LEDs.

It uses Earle Philhower's RP2040 Arduino board package, but should be easy to adapt for the Arduino package when it becomes available.


ladyada (approved):

ladyada:
thanks!

PR #275, opened 2021-05-28
Fix Arduino Nano33BLE with core >= 2.0.0

facchinm:
Latest mbed-based cores hide the porting layer to greatuly speedup compilation.
In case direct access to such structures is needed, pinDefinitions.h must be included.

By checking the file presence via __has_include we also keep compatibility with previous version of the core

Fixes https://github.com/arduino/ArduinoCore-mbed/issues/241


ladyada:
thnkx :)

Issue #278, opened 2021-06-29
Error compiling for board ESP32C3 Dev Module


LilyGO:
How does the NeoPixel library set the **ESP32_C3** development board environment?
Compilation fails with the basic example.

ladyada:
theres no support - please add it by submitting a PR!

patlkli:
This can be closed then, right?

LilyGO:
ok

PR #279, opened 2021-06-29
Fix build for ESP32-C3

patlkli:
This PR fixes the build issues for the ESP32-C3 Dev Module reported in #278.

The newest development release of the ESP32 Arduino core (2.0.0-alpha1) brings support for the ESP32-C3.
It also integrates ESP-IDF v4.4, which finally brings the missing functions in the RMT driver that were previously workarounded in `esp.c`.
Since the clock selection seems to only have gotten more complicated and stuff like `RMT_BASECLK_REF` is not even available for the ESP32-C3 target.
So to avoid pulling any more complexity from ESP-IDF in here, I think the easiest fix is to just use the IDF's own functionality when compiling against a new enough version.

### DIFF
--- a/esp.c
+++ b/esp.c
@@ -97,6 +97,10 @@ void espShow(uint8_t pin, uint8_t *pixels, uint32_t numBytes, boolean is800KHz) 
         return;
     }
 
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4, 0, 0)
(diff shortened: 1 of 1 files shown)


ladyada:
thanks! didnt test it

LilyGO:
I tested the code yesterday and it worked perfectly. Thank you so much

PR #291, opened 2021-09-15
Fix build with older ESP-IDF versions

patlkli:
This PR fixes the build errors with ESP-IDF versions before v4.x that come with arduino-esp32 versions before v2.x.
The errors were caused by `ESP_IDF_VERSION` and `ESP_IDF_VERSION_VAL` not being available in ESP-IDF v3.x.

I'm not sure how I missed this obvious oversight in my previous PR, but I this should fix it.

I tested this one with the following boards and arduino-esp32 versions.

- ESP32-DevKitS + external WS2812 (arduino-esp32 v1.0.4-1.0.6 and v2.0.0)
- ESP32-S2-Saola-1 + on-board WS2812 (only supported since arduino-esp32 v2.0.0)
- ESP32-C3-DevKitC-02 + on-board WS2812 (only supported since arduino-esp32 v2.0.0)

### DIFF
--- a/esp.c
+++ b/esp.c
@@ -22,6 +22,12 @@
 #include <Arduino.h>
 #include "driver/rmt.h"
 
+#if defined(ESP_IDF_VERSION)
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4, 0, 0)
+#define HAS_ESP_IDF_4
+#endif
+#endif
+
 // This code is adapted from the ESP-IDF v3.4 RMT "led_strip" example, altered
 // to work with the Arduino version of the ESP-IDF (3.2)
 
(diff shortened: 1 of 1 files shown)


ladyada (approved):

ladyada:
yay thanks! ok will merge after it passes CI

patlkli:
BTW I just checked why my old PR even passed CI.

Reason is that `ci-adafruit` was already using the `dev` board manager url (since [this commit](https://github.com/adafruit/ci-arduino/commit/31575a9335032df77031292500f4cf9096f5d992)).
At the time, that already had the alpha version of arduino-esp32 v2.0.0 version in it, which doesn't exhibit this error.

Now arduino-esp32 v2.0.0 is out proper, but back when I did the other PR, CI was testing against an alpha version of the ESP32 Arduino framework.

PR #306, opened 2021-12-17
Update RP2040 code to find a free PIO state machine

Bodmer:
The original RP2040 PIO and state machine choice was hard coded into the library. This change will search for and use a free PIO state machine.

Typo also fixed and version incremented.

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=Adafruit NeoPixel
-version=1.10.1
+version=1.10.2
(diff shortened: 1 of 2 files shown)


Bodmer:
Does not work with more than 1 string.

dhalbert:
@Bodmer Why did you close the PR?

Bodmer:
See #307, for an improved update.

PR #312, opened 2022-01-21
Add support for SAMD11C14

ottobonn:
This change just adds a check for the SAMD11C14 microcontroller, which uses the same code section as its siblings to write to the pixels. 

I have tested with an ATSAMD11C14 running on a custom PCB with RGBW NeoPixels. I'm using MattairTech's generic ATSAMD11C14 core and PA04 (pin 4) as the LED data pin.

Please let me know if there are any other things I need to address. Without this change, the ATSAMD11 doesn't have any code in the `show` function, so the LEDs do nothing.

Resolves https://github.com/adafruit/Adafruit_NeoPixel/issues/157

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -2253,10 +2253,11 @@ void Adafruit_NeoPixel::show(void) {
   }
   // END of NRF52 implementation
 
-#elif defined(__SAMD21E17A__) || defined(__SAMD21G18A__) ||                    \
-    defined(__SAMD21E18A__) ||                                                 \
-    defined(__SAMD21J18A__) // Arduino Zero, Gemma/Trinket M0, SODAQ Autonomo
-                            // and others
+#elif defined(__SAMD21E17A__) || defined(__SAMD21G18A__) || \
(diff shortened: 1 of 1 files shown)


ladyada:
i can dig it! please hiya! thanks so much for submitting a PR! we can review & merge PRs once they have passed continuous integration (CI). that means that code compiles cleanly for all tested platforms, is clang formatted so we maintain the same text formatting for all new code, and is doxygen documented.
if your code isnt passing, check the CI output (click on the red X next to the PR to scroll through the log and find where the error is

here is a tutorial on doxygen: https://learn.adafruit.com/the-well-automated-arduino-library/doxygen

and clang-format: https://learn.adafruit.com/the-well-automated-arduino-library/formatting-with-clang-format

and overall how to contribute PRs to libraries: https://learn.adafruit.com/contribute-to-arduino-with-git-and-github

once you get that green checkmark that indicates CI has passed, please comment reply to this post so we know its time for another review (we may not get notified on CI pass and miss that its time to look!)

ottobonn:
Looks like CI has passed!

ladyada (approved):

ladyada:
woot!

PR #324, opened 2022-07-06
added setPixelColor(uint16_t n, uint32_t c, uint8_t brt) function

moose4lord:
I've created a project which is more focused on individual LED brightness, rather than color, so this new setPixelColor() function makes manipulating the LEDs a bit easier. It sets an LED's brightness directly, and ignores the global strip brightness.

Since it's a new function, I don't think it should break any existing code. Here is my test code, which fades a strip of LEDs in and out:
```c++

#define PIN      12
#define NUMPIXELS 8

Adafruit_NeoPixel pixels(NUMPIXELS, PIN, NEO_GRB + NEO_KHZ800);

uint8_t brightness = 0;
int8_t dir = 1;

void setup() {
  pixels.begin();
}

void loop() {
  for(int i=0; i<NUMPIXELS; i++) { // fill strip with BLUE at a given brightness
    pixels.setPixelColor(i, 0x0000ff, brightness);
  }
  brightness += dir; // increase or decrease brightness for each loop

  if(brightness == 0 || brightness == 255) { // change fade direction
    dir *= -1;
  }

  pixels.show();
  delay(10);
}
```

Note, the existing setPixelColor(uint16_t n, uint32_t c) could be optimized to this:
```c++
void Adafruit_NeoPixel::setPixelColor(uint16_t n, uint32_t c) {
  Adafruit_NeoPixel::setPixelColor(n, c, brightness - 1);
}
```
But I did not do that in case you prefer readability over optimization.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -3153,6 +3153,38 @@ void Adafruit_NeoPixel::setPixelColor(uint16_t n, uint32_t c) {
 }
 
 /*!
+  @brief   Set a pixel's color using a 32-bit 'packed' RGB or RGBW value. Sets
+           the color brightness directly, ignoring the global strip brightness.
+  @param   n  Pixel index, starting from 0.
+  @param   c  32-bit color value. Most significant byte is white (for RGBW
+              pixels) or ignored (for RGB pixels), next is red, then green,
+              and least significant byte is blue.
+  @param   brt  Brightness setting, 0=minimum (off), 255=brightest.
+*/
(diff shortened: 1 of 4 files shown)


moose4lord:
The automated CI checks caught a couple syntax errors in the strandtest_nodelay and StrandtestBLE_nodelay example sketches (unrelated to the addition of the new setPxelColor() function). They were simple fixes, so I threw those in as well.

moose4lord:
I'm withdrawing this PR. After more testing and evaluation I find it would be more useful to have two simpler functions that just get/set the raw pixel data, bypassing any brightness calculations altogether. I'll try to resubmit a PR with that in mind.

PR #330, opened 2022-09-21
Fixed: arduino-esp32 2.0.5 calling abort() from espShow()

rimim:
- Disabled calls to noInterrupt()/interrupt() in espShow() on ESP32. espShow() calls rmt_driver_install() which tries to acquire a mutex. Recursive mutexes are not allowed inside an interrupt context so lock_acquire_generic() calls abort().

- Change only applies to ESP32 platforms and the calls to noInterrupt()/interrupt() are not needed as RMT will handle all the bit twiddling.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -283,7 +283,8 @@ void Adafruit_NeoPixel::show(void) {
     // to the PORT register as needed.
 
     // NRF52 may use PWM + DMA (if available), may not need to disable interrupt
-#if !(defined(NRF52) || defined(NRF52_SERIES))
+    // ESP32 may not disable interrupts because espShow() uses RMT which tries to acquire locks
+#if !(defined(NRF52) || defined(NRF52_SERIES) || defined(ESP32))
   noInterrupts(); // Need 100% focus on instruction timing
 #endif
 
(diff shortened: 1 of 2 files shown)


rimim:
arduino-esp32 2.0.4 defines interrupts() and noInterrupts() as no-ops.

    #define sei()
    #define cli()
    #define interrupts() sei()
    #define noInterrupts() cli()

arduino-esp32 2.0.5 calls portENABLE_INTERRUPTS() and portDISABLE_INTERRUPTS().

    // ESP32xx runs FreeRTOS... disabling interrupts can lead to issues, such as Watchdog Timeout
    #define sei() portENABLE_INTERRUPTS()
    #define cli() portDISABLE_INTERRUPTS()
    #define interrupts() sei()
    #define noInterrupts() cli()

This is why the NeoPixel library is crashing (calling abort) when compiled with 2.0.5. It's also why it should be valid to remove the calls to noInterrupt()/interrupt() as they never did anything on the ESP32 previously.

brentru:
@rimim I also experienced this same issue with 2.0.5 and WipperSnapper, and traced it back to NeoPixel. Thanks for submitting a fix!

I've successfully tested this on a Feather ESP32 v2 running WipperSnapper (which shows the status of WiFi connectivity via the NeoPixel)

brentru (approved):

ladyada (approved):

PR #365, opened 2023-09-16
Add initial support for Arduino UNO R4 boards

KurtE:
The code is sort of a cross between several of the other ARM boards, like the Teensy T4 boards, plus setting up registers using code similar to what I did with the Servo library. Plus some other defines from what @facchinm did in the FastLED library to gain access to the ARM registers

Code was added in a new 
#elif defined(ARDUINO_ARCH_RENESAS) || defined(ARDUINO_ARCH_RENESAS_UNO) || `defined(ARDUINO_ARCH_RENESAS_PORTENTA) 
`
section in the source file for the new boards, 

So it should not impact other boards.

I was also tempted to throw in a #else section before ending the #if for ARM is:

```
#endif // ARM

  // END ARM ----------------------------------------------------------------

```
Probably like:
```
#else
#error Architecture not supported
#endif // ARM

```
So that other new ARM based boards would not compile without any errors or warnings, but not have any actual code to
implement them.

But decided that was a different issue

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -2914,7 +2914,89 @@ if(is800KHz) {
     ; // Wait for last bit
   TC_Stop(TC1, 0);
 
-#endif // end Due
+
+// RENESAS including UNO R4
+#elif defined(ARDUINO_ARCH_RENESAS) || defined(ARDUINO_ARCH_RENESAS_UNO) || defined(ARDUINO_ARCH_RENESAS_PORTENTA)
+
+// Definition for a single channel clockless controller for RA4M1 (Cortex M4)
+// See clockless.h for detailed info on how the template parameters are used.
+#define ARM_DEMCR               (*(volatile uint32_t *)0xE000EDFC) // Debug Exception and Monitor Control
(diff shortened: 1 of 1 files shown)


F-Schmidt99:
Dear Maintainers,

Could you please review and consider this merge request for adding NeoPixel library support for UNO R4? Existing libraries (or their forks to be exact) lack a seamless installation option via the Arduino IDE Library Manager, and this addition would greatly enhance user convenience.

Your approval and inclusion in the next library update would be highly appreciated.

Thank you for your time.

metehoca:
Can't get this code to work. Trying "simple" example, changed lots of pins but no luck. Any ideas? R4 Minima.

ladyada:
ill merge but its not tested, would be great if someone can verify this works

F-Schmidt99:
Setup:
- Arduino UNO R4 Minima
- 1x WS2813 connected to Pin D3 of Arduino
- Software: Adafruit Neopixel in its current version downloaded and manually installed from this repo.
- Program: Example-Sketch 'simple.ino'

Result:
The LED show very bright white color. `setBrightness(...)` or `setColor(...)` do not show any effect.
**Not working.**

PR #367, opened 2023-10-28
Espidf5 new RMT support

ladyada:
cc @me-no-dev & https://github.com/espressif/arduino-esp32/issues/8796

i think this is right? i dont need to disable the rmt post-write and it's ok that the `rmt_data_t led_data[numBytes * 8]` its writing is on the function stack? or should i malloc it on the heap?

### DIFF
--- a/esp.c
+++ b/esp.c
@@ -26,7 +26,21 @@
 #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4, 0, 0)
 #define HAS_ESP_IDF_4
 #endif
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
+#define HAS_ESP_IDF_5
 #endif
+#endif
+
+
+
+#ifdef HAS_ESP_IDF_5
+
+
+
+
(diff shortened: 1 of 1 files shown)


me-no-dev:
@ladyada I am not sure how many `bytes` this function will try to send, but each `byte` takes 32 actual bytes of RMT memory. Maybe best to malloc if data might grow. Also I see you test against IDF version, but at the same time you use Arduino API. Why not test against the Arduino version? We have a header since early 2.x and is much like IDF's :)

me-no-dev:
Oh and on the disabling RMT, no you do not need to do that. You could even later use that same pin for something else and as soon as you try to use it for it, the peripheral manager in Arduino will deinit the RMT and init whatever you asked.

me-no-dev:
> i think testing against IDF instead of BSP might be better because some people could use this code in IDF directly

It calls ESP32 Arduino API's so even if used in IDF, Arduino needs to be added as component. It generally does not matter which one you use, at least for now they mean the same thing, but in the future that might not be the case. I'm merely suggesting it :)

PR #370, opened 2023-11-05
Added code for static arrays

philm001:
Added in provision to incorporate the use of static arrays for LEDs. I was working on a project that used 11 different strips of Neo pixels (this was the  144 LED/m where it is recommended to use only 1 meter at a time). Since I used 11 strips, I needed 11 different instances of the Adafruit class and with this being dynamic memory on a MCU, this killed my heap. Even the ESP32 had a hard time.

But, once I removed the dynamic aspect from the library, everything worked. This code will be valuable to those that use more then 1 strip. I also have default arguments to be backward compatible with existing code.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -77,10 +77,11 @@
               pixel.
   @return  Adafruit_NeoPixel object. Call the begin() function before use.
 */
-Adafruit_NeoPixel::Adafruit_NeoPixel(uint16_t n, int16_t p, neoPixelType t)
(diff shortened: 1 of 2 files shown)


philm001:
So any update on this......

dhalbert:
It looks like the storage is allocated in the constructor and deallocated in the destructor. Are you constructing/destructing over and over again, or just once and your ran out of heap? Did you shared the buffer between multiple instances to avoid that?

It would be good to pass the length of the buffer and validate that it's at least as big as what's required.

philm001:
Hello, thank you for getting back. I haven't look at the code in awhile so will need to check and get back to you

PR #381, opened 2024-02-24
rp2040 check for available pio instruction memory

bbacskay:
Hi All,

On the RP2040(W) the class caused a PANIC when I tried to use it together with another library that was also using the PIO. I initialized the other library first and it happened to use all of the instruction memory of pio0. The original code just checked for available SM, but not for pio memory.

I changed the rp2040 init code to check for available memory first on pio0, if that fails then it tries to load the pio program to pio1's memory and if that also fails then the raspberry SDK throws PANIC.
I also added clean up code to the destructor to free the pio resources.

This is a compatible change and effects only the rp2040 devices.

Probably already known limitation:
With the current pio program one rp2040 can only handle 8 Neopixel strips and the pio program is going to be loaded separately for each instance.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -133,6 +133,10 @@ void Adafruit_NeoPixel::begin(void) {
   if (pin >= 0) {
     pinMode(pin, OUTPUT);
     digitalWrite(pin, LOW);
+
+    #if defined(ARDUINO_ARCH_RP2040)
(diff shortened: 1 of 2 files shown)


ladyada:
thank you! i hand-implemented this in 
https://github.com/adafruit/Adafruit_NeoPixel/pull/436
and added RP2350 pio2 support at the same time
your code was helpful :)

PR #388, opened 2024-05-02
Add files via upload

iv-vi:
Fixed 8bit AVR uc issue with assembler code for PORTF.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -823,11 +823,11 @@ void Adafruit_NeoPixel::show(void) {
       asm volatile(
           "headF:"
           "\n\t"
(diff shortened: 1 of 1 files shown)


Adafruit_NeoPixel.cpp
```diff
@@ -2914,89 +2917,7 @@ if(is800KHz) {
...
-        }
-        *clr = msk;
-      }
-    }
-    while (ARM_DWT_CYCCNT - cyc < CYCLES_400)
-      ;
-  }
-#endif // NEO_KHZ400
-
-#endif // ARM
```
razvanphp, line 2999: Why would you remove this?!

This code works beautifully.

ladyada:
hiya - we won't review until renesas is un-deleted

Issue #375, opened 2023-12-17
pixel.show() crash with more than 73 pixel on ESP32s3


tresler:
I have weird problem. I test [example code](https://github.com/adafruit/Adafruit_NeoPixel/blob/master/examples/simple/simple.ino) or [example with new operator](https://github.com/adafruit/Adafruit_NeoPixel/blob/master/examples/simple_new_operator/simple_new_operator.ino) and everything works fine if I have 73 or less RGB leds. If I use 74 or more, everything crash. Can anyone explain it?

Debug code: [adafruit_neopixel_crash.txt](https://github.com/adafruit/Adafruit_NeoPixel/files/13697955/adafruit_neopixel_crash.txt)

- Arduino board:  ESP32s3-WROOM
- Arduino-ESP32: 3.0.0 alpha 3
- Arduino IDE: 2.2.1
- Adafruit_Neopoxel: 1.12

ajkeeton:
Same problem here, also Esp32 (Heltec Wifi Kit), and at around 70 pixels

teknynja:
I'm experiencing the same issue as well (crash/boot loop when driving more than around 70 pixels). This occurs on both my XIAO ESP32C3 boards and my ESP-WROOM-32 Dev Modules, with the following hardware/software:

* Boards: XIAO ESP32C3, Generic ESP-WROOM-32 38 Pin Dev Board
* Board Package: esp32 by Espressif (both v3.0.0 & v3.0.1)
* Adafruit NeoPixel v1.12.2
* Arduino IDE: 2.3.2

[...]

PR #392, opened 2024-06-09
RMT Buffer Allocation Fix for Issue #375

teknynja:
**_NOTE: This pull requests precedes pull request #394 - I've since added a mutex for thread safety, I'm leaving this in place so both versions are available for evaluation._**

This pull request addresses issue #375 [pixel.show() crash with more than 73 pixel on ESP32s3]

After reviewing the code and also getting some helpful feedback from the Espressif Forums (https://www.esp32.com/viewtopic.php?f=13&t=40270) and @robertlipe it was determined that the code in esp.c for handling the RMT item buffers when using the IDF v5 framework was allocating too much space on the stack when using more than 70ish pixels.

This pull request addresses that issue by allocating the RMT buffers from the heap instead. It will attempt to allocate a single block of memory to accommodate the largest configured instance (sharing the buffer between instances is fine, as the buffer is completely populated each each time the `espShow()` method is called).

I also took the time to improve the channel allocation management, previously the RMT channels were initialized on each call to `espShow()`, now the RMT channels are only de-initialized and re-initialized whenever the output pin is changed.

Finally, I was concerned about problems that may be caused by allocating large buffers on the heap without giving the user any way to free that memory, so the code allows a user to free that memory (and also release the RMT channels) by setting the number of pixels to zero using `.updateLength(0)` and then calling `.show()`. This will de-allocate the heap memory used for the RMT buffers and release the RMT channels held by driver. They will automatically be re-allocated if needed when setting the number of pixels back to a non-zero value.

It seems one of the primary goals on this project is to minimize the distribution changes across files, to that end you'll find that all the changes here are constrained to a single method in the esp.c file. A different approach would be required to address thread-safety and instance sharing issues, but it seems like there is existing code that is problematic in those areas as well so those goals were not prioritized. I think for a majority of the users of this library, those are not going to be issues anyways.

### DIFF
--- a/esp.c
+++ b/esp.c
@@ -35,33 +35,72 @@
 #ifdef HAS_ESP_IDF_5
 
 void espShow(uint8_t pin, uint8_t *pixels, uint32_t numBytes, boolean is800KHz) {
-  rmt_data_t led_data[numBytes * 8];
+  // Note: Not thread-safe!!! Also because rmtPin is shared between
+  //  all instances, we will end up releasing/initializing the RMT
+  //  channels each time we invoke on different pins. This is probably
+  //  ok, just not efficient. led_data will be allocated with enough
+  //  space for the largest instance but data is not used between
+  //  calls so this should be fine.
+  static rmt_data_t *led_data = NULL;
+  static uint32_t led_data_size = 0;
+  static int rmtPin = -1;
 
-  if (!rmtInit(pin, RMT_TX_MODE, RMT_MEM_NUM_BLOCKS_1, 10000000)) {
-    log_e("Failed to init RMT TX mode on pin %d", pin);
+  uint32_t requiredSize = numBytes * 8;
+  if (requiredSize > led_data_size) {
+    free(led_data);
+    if (led_data = (rmt_data_t *)malloc(requiredSize * sizeof(rmt_data_t))) {
+      led_data_size = requiredSize;
+    } else {
+      led_data_size = 0;
+    }
+  } else if (requiredSize == 0) {
+    // To release RMT resources (RMT channels and led_data), call
+    //  .updateLength(0) to set number of pixels/bytes to zero,
+    //  then call .show() to invoke this code and free resources.
+    free(led_data);
+    led_data = NULL;
+    if (rmtPin >= 0) {
+      rmtDeinit(rmtPin);
+      rmtPin = -1;
+    }
+    led_data_size = 0;
     return;
   }
 
-  int i=0;
-  for (int b=0; b < numBytes; b++) {
-    for (int bit=0; bit<8; bit++){
-      if ( pixels[b] & (1<<(7-bit)) ) {
-        led_data[i].level0 = 1;
-        led_data[i].duration0 = 8;
-        led_data[i].level1 = 0;
-        led_data[i].duration1 = 4;
-      } else {
-        led_data[i].level0 = 1;
-        led_data[i].duration0 = 4;
-        led_data[i].level1 = 0;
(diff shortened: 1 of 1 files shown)


teknynja:
Cool, thanks!

As for the free/malloc vs realloc, I just copied how the `pixels` buffer is allocated (when in Rome... :smirk:).

teknynja:
I completely agree with you on these safety issues. There are a few different areas of consideration that come to mind regarding this:

* RMT Channel Management – The Espressif documentation on the new RMT API is currently MIA, and as a placeholder they are using a bit of example code (In fact it is for driving NeoPixels, and looks suspiciously like the code I found in the Adafruit_NeoPixel espShow method 😏). I believe the intent of this API is to hide all the messy channel allocation/management code and to deal with all the buffering of data through those channels. In fact, I’m currently using it with two different “clients” in the same sketch, one being the Adafruit_NeoPixel library, and the other an IR Driver I wrote (because I couldn’t find a library that supports Arduino Core v3 and used the RMT hardware). Both libraries are working great together on an S3 chip (which has 8 read/write RMT channels) and a C3 chip (which has 2 read and 2 write channels with smaller buffers). The RMT API seems to be doing its job and keeping the channels separate and full.

* Instance Management – Because the ESP32 code in the Adafruit_NeoPixel library is not a “first-class citizen” of the library, it doesn’t have access to instance-level data; it can only access ephemeral data on the stack or global data shared across all instances of the library. The only fix for this is to make the code in the esp.c module part of the Adafruit_NeoPixel class, which seems like something a project lead would need to be involved with. This could also potentially introduce some negatives, for example each instance might allocate heap for it’s own RMT symbol buffer, which could possibly eat up a large portion of memory. Another consideration are the functions that should be in IRAM, which makes pulling them into “class” code more challenging.

* Thread Safety – This is arguably the harder problem to deal with. To date it looks like (other than enabling/disabling interrupts) only the NRF52 code is thinking about thread safety (and mostly just the use of `rtos_malloc`, which might be a benefit everywhere?). I’m pretty sure we can’t disable interrupts around the RMT code (it’s probably using them to pump data through the channels), but a mutex might help protect some of the critical code here. A quick search seems to point to several different ways (pthreads, multiple-cores, FreeRTOS xTaskX, etc) to implement multi-threading on an ESP32 Arduino project, all of which need to be considered. Again, this feels like project-lead level stuff here, and should probably be looked at by someone with more Arduino multi-threading experience than I have. 

All this being said, I think the currently proposed code does the best it can given then current environment, and at least allows users the ability to drive larger numbers of pixels when building against Core v3 (of course, with caveats). Properly addressing the above issues will likely require a fair bit of restructuring of the library code, and feels out-of-scope for fixing the actual problem users are currently experiencing. When the above issues are addressed, all of the ESP32 code (including the old Core v2 code) will need to be re-worked to bring it up to the level of safety we are considering here.

teknynja:
OIC - I wasn't sure about your relationship with this project, but I guess if you hadn't come along to encourage me (and the maintainers stayed quiet) then I probably wouldn't have tried to tackle this.

This is essentially my first pull request for an open-source project, so I know nothing about the protocol. I kind of thought that simply pushing additional changes to my fork would end up in this pull request. Would it be confusing for the maintainer to have two separate forks/pull requests from the same person for the same issue? TBH the changes to add the mutex weren't all that invasive (but it did change the code flow a bit).

I see your point about the array size, I don't do a lot of c/c++ coding (Internet to the rescue!), but I will keep that in mind for the future!

dhalbert:
I'm going to close this in favor of #394, at least for now.

PR #397, opened 2024-07-03
add ch32 support

hathach:
add support for ch32 core with bit banging implementation. Tested with ch32v203g6 qtpy with neopixel featherwing (32 pixel) work with multiple F_CPU: 144mhz, 120mhz, 96mhz, 72mhz, 56mhz

@ladyada


ladyada:
yay thanks!

Issue #359, opened 2023-06-17
Unused variable warning in SAMD51 implementation


IanBurwell:
When compiling for the PyPortal (SAMD51) GCC raises an unused variable warning (see below). This issue is not critical but would be easily fixed by removing the hanging `bit` variable [here](https://github.com/adafruit/Adafruit_NeoPixel/blob/fe882b84951bed066764f9350e600a2ec2aa5a9e/Adafruit_NeoPixel.cpp#LL2505C44-L2505C47)

GCC (9.3.1) warns the following:
```bash
.pio/libdeps/adafruit_pyportal_m4_titano/Adafruit NeoPixel/Adafruit_NeoPixel.cpp: In member function 'void Adafruit_NeoPixel::show()':
.pio/libdeps/adafruit_pyportal_m4_titano/Adafruit NeoPixel/Adafruit_NeoPixel.cpp:2505:44: warning: unused variable 'bit' [-Wunused-variable]
 2505 |   uint8_t *ptr, *end, p, bitMask, portNum, bit;
      |                                            ^~~
```

PR #409, opened 2024-10-29
Removed unused `bit` variable for SAMD51

Grochocinski:
## Scope of Change
Removed unused `bit` variable for SAMD51 microcontrollers in `Adafruit_NeoPixel.cpp`

## Limitations of Change
Change only applies to SAMD51 microcontrollers

## Other Notes
Fixes issue #359, PR #235 was previously opened to fix this but was self-closed


ladyada:
thank you :)

PR #428, opened 2025-04-09
Add support for Arduino Portenta H7 and Giga R1 boards

razvanphp:
Inspired from cf41ed1 , Thank you @KurtE!

Successfully tested on Arduino Portenta H7 on the bench.


ladyada:
thanks, i'm trusting that this was tested and has no side effects. may be reverted if found otherwise!

PR #431, opened 2025-04-15
FIX: ESP32 compatibility 

karshPrime:
# Scope of Change:
This pull request addresses a compile-time error encountered in the library when used with the ESP-IDF. The specific issue arises from the use of an assignment within an `if` statement, which results in a warning being treated as an error by the compiler.

The original line of code:
```c
if (led_data = (rmt_data_t *)malloc(requiredSize * sizeof(rmt_data_t))) {
```
has been modified to separate the assignment from the conditional check. The updated code is as follows:
```c
led_data = (rmt_data_t *)malloc(requiredSize * sizeof(rmt_data_t));
if (led_data) {
```
This change ensures that the memory allocation is checked properly, thereby preventing potential runtime errors if the allocation fails.

**I have also added `**/build` to `.gitignore`, as this directory is generated when building the project with `idf.py build`.**

## CMake
NOT ADDED IN THIS PR, but I believe a root `CMakeLists.txt` file could be created with similar configuration as the following for `$ idf.py build`:
```
idf_component_register(
    SRCS
        "Adafruit_NeoPixel.cpp"
        "esp.c"
        "esp8266.c"
        "kendyte_k210.c"

    INCLUDE_DIRS
        "."

    REQUIRES
        "arduino"
)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 Doxyfile*
 doxygen_sqlite3.db
 html
+**/build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,13 @@
+idf_component_register(
+    SRCS
+        "Adafruit_NeoPixel.cpp"
+        "esp.c"
+        "esp8266.c"
+        "kendyte_k210.c"
+
+    INCLUDE_DIRS
+        "."
+
+    REQUIRES
+        "arduino"
+)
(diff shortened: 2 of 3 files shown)


karshPrime:
I am unsure whether the `CMakeLists.txt` file might lead to any issues with other project types; therefore, it has not been included in this pull request.

dhalbert:
What version of the arduino-esp32 BSP or ESP-IDF are you using? We test the libraries by compiling them on various platforms supplied for Arduino.

karshPrime:
I'm getting the following error with IDF v5.1.5.

image

PR #435, opened 2025-05-20
Fixed NRF51 implementation (BBC Microbit)

### LINKED ISSUE #374: No pwm output on micro:bit pin 8
- Arduino board:  **BBC Micro:bit v1**

- Arduino IDE version (found in Arduino -> About Arduino menu):  2.2.1 (Also using VS Code with the Arduino extension

- List the steps to reproduce the problem below (if possible attach a sketch or
  copy the sketch code in too):

```  

const int numPixels = 24;
const int pixelPin = 8;

Adafruit_NeoPixel ring = Adafruit_NeoPixel(numPixels, pixelPin);
uint32_t foreground = ring.Color(0x00, 0x00, 0xFF); // r, g, b - blue
uint32_t background = ring.Color(0x10, 0x10, 0x10); // dim white

void setup() {
  pinMode(pixelPin, OUTPUT);
  ring.begin(); // start the NeoPixel display
}

void loop() {
  // blue dot circles around a white background (for PixelRing 24)
  for (int i = 0; i < numPixels; i++) {
    ring.setPixelColor(i, foreground); // set pixel i to foreground
    ring.show();                       // actually display it
    delay(50);                         // milliseconds 
    ring.setPixelColor(i, background); // set pixel to background before moving on
    ring.show(); 
  }
}
```
Using pin 8 as the LED pin does not work. There's no output at all on the pin. Changing to pin 2 results in a working sketch. Pin 8 is definitely capable of pwm (tested with the below sketch).
```
const int LED = 8;

void setup() {   
  pinMode(LED, OUTPUT);   
}

void loop(){
  analogWrite(LED, 128);
  delay(1000);
  analogWrite(LED, 1);
  delay(1000);
}
```

mattybigback:
Fixes #374

The NRF51 implementation did not work when using higher pin numbers.  strb was being used, which only stores the least significant byte. This prevented higher pin numbers from being used. Modified code to use str.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -2988,7 +2988,8 @@ if(is800KHz) {
   unsigned int bitmask = (1 << g_ADigitalPinMap[pin]);
   // https://github.com/sandeepmistry/arduino-nRF5/blob/dc53980c8bac27898fca90d8ecb268e11111edc1/variants/BBCmicrobit/variant.cpp
 
-  volatile unsigned int *reg = (unsigned int *)(0x50000000UL + 0x508);
+  //volatile unsigned int *reg = (unsigned int *)(0x50000000UL + 0x508);
+  volatile uint32_t *reg = (uint32_t *)(NRF_GPIO_BASE + offsetof(NRF_GPIO_Type, OUTSET));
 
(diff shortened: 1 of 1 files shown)


mattybigback:
Follow up comment to add testing method carried out.

`/examples/strandtest/strandtest.ino` was tested with a 60 LED ring and a BBC Microbit. For each iteration, `LED_PIN` was set to every available GPIO pin (0-17, 19-20). A full cycle of the loop was observed for each pin.

dhalbert (approved):
Thank you!

PR #436, opened 2025-05-24
clean up and hand-add some of https://github.com/adafruit/Adafruit_NeoPixel/pull/381

ladyada:
some work on rp2350, add support for pio2 and make it fail quietly rather than hanging


ladyada:
thanx!

Issue #437, opened 2025-06-01
Compilation error for Arduino Nano RP2040 Connect


humzakh:
Hi, I'm getting a compilation error using Adafruit_NeoPixel 1.15.1. Any help would be greatly appreciated.

```
/Users/hkhan/Documents/Arduino/libraries/Adafruit_NeoPixel/Adafruit_Neopixel_RP2.cpp: In member function 'bool Adafruit_NeoPixel::rp2040claimPIO()':
/Users/hkhan/Documents/Arduino/libraries/Adafruit_NeoPixel/Adafruit_Neopixel_RP2.cpp:9:9: error: 'pio_claim_free_sm_and_add_program_for_gpio_range' was not declared in this scope
   if (! pio_claim_free_sm_and_add_program_for_gpio_range(&ws2812_program,
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/hkhan/Documents/Arduino/libraries/Adafruit_NeoPixel/Adafruit_Neopixel_RP2.cpp: In member function 'void Adafruit_NeoPixel::rp2040releasePIO()':
/Users/hkhan/Documents/Arduino/libraries/Adafruit_NeoPixel/Adafruit_Neopixel_RP2.cpp:35:3: error: 'pio_remove_program_and_unclaim_sm' was not declared in this scope
   pio_remove_program_and_unclaim_sm(&ws2812_program, pio, pio_sm,  pio_program_offset);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/hkhan/Documents/Arduino/libraries/Adafruit_NeoPixel/Adafruit_Neopixel_RP2.cpp:35:3: note: suggested alternative: 'pio_remove_program'
   pio_remove_program_and_unclaim_sm(&ws2812_program, pio, pio_sm,  pio_program_offset);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   pio_remove_program
exit status 1

Compilation error: exit status 1
```

je1rav:
The same error occurred on "Arduino Mbed OS RP2040 Boards by Arduino"

caternuson:
The Arduino Mbed OS Nano Board Package is using an older version of pico-sdk:
[...]

PR #439, opened 2025-06-27
Fix pio related compilation issues for BSP's with older Pico SDKs

caternuson:
Potential fix for #437 

Updates preproc to check for Pico SDK major version and only support pio usage if at least 2.x.

Functional check on a Metro RP2350 and a 12 pixel ring on D2 using `strandtest` example sketch.

Only did a compilation check for Arduino Nano RP2040 Connect.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -126,7 +126,7 @@ Adafruit_NeoPixel::~Adafruit_NeoPixel() {
 #endif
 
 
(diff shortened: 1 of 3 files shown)


ladyada:
is this to fix for people using the arduino BSP?

ladyada:
ok i mean, it will compile but wont output anything on the old BSP because the PIO is never configured (?)

ladyada:
isnt it better we dont compile then, rather than just silently not working?

caternuson:
closing this. sorry for the noise.

PR #443, opened 2025-08-04
esp: fix -Wparenthesis warning

zachwelch:
This trivial patch fixes the following warning:
```
.../Adafruit_NeoPixel/esp.c: In function 'espShow':
.../Adafruit_NeoPixel/esp.c:56:11: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
   56 |       if (led_data = (rmt_data_t *)malloc(requiredSize * sizeof(rmt_data_t))) {
      |           ^~~~~~~~
```


ladyada:
thanks!

Issue #438, opened 2025-06-15
Last working Arduino Nano BLE version 1.13.0


tickietackie:
Using the standard example strand test from the library on port 6 works with my Arduino Nano BLE and second one only until version 1.13.0, for version 1.14.0 and higher it's not working anymore, the LEDs are not showing anything. Downgrading the library reliantly resolves the issue and also my other projects in Platform.io then do work again with the same method.

Do I miss something here or changed sth. in the library that bricked the Nano BLE?

jmcenteeIV:
We ran into this same issue.  I am posting a blurb that our EE wrote up about the issue in the hopes it may be helpful in identifying the problem:

> "The WS2812b 5050 LED strips appear to be incompatible with the Arduino 33BLE REV2 processor when using Adafruit NeoPixel library version 1.15.1.  The programmed D2 output of the Arduino exhibited approximate 100KHz serial data signal versus the expected 800KHz signal.  The LEDs on the strip were not addressed properly and all lights were erroneously turned on with full white brightness..  
When the library was switched to the older version 1.13.0, the expected 800KHz signal was present and the LED strip was addressed properly.  
LED strip products continue to evolve beyond the WS2812b with FCOB (Flip Chip On Board) and others, so there appears to be compatibility issues that are becoming more prevalent."

PR #447, opened 2025-09-22
Issue #438: Fixes for Arduino Nano 33 BLE.

chuckstanski:
This PR removes code that was intended to support the Arduino Giga R1 boards; however, its integration into the codebase broke the Arduino Nana 33 BLE devices and possibly other ARM M4 device.  The problematic macro was the use of `defined(TARGET_M4)` which unintentionally assumed all ARM M4 chips are using a 240MHz clock.  The Nano 33 BLE boards have a ARM M4 at their core but run at a much lower clock rate.  In addition the Nano 33 BLE boards use a different mechanism to control the GPIO (see NRF52 code in Adafruit_NeoPixel.cpp starting at line 2169) which was not used due to the mentioned macro.  

Arduino Giga R1 support is already in the codebase (see Adafruit_NeoPixel.cpp line 3136 and other lines) which the macro `defined(TARGET_GIGA) || defined(TARGET_M4)` bypassed due to its placement in the file as well.

With the proposed changes, the Arduino Nano 33 BLE board has been tested.  The Arduino Giga R1 code has been verified to compile correctly and include the correct code (i.e. `defined(ARDUINO_ARCH_MBED_GIGA)`); however, has not been verified on target due to lack of a Giga R1 board.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -3083,9 +3083,9 @@ if(is800KHz) {
 #define ARM_DWT_CTRL_CYCCNTENA          (1 << 0)                // Enable cycle count
 #define ARM_DWT_CYCCNT          (*(volatile uint32_t *)0xE0001004) // Cycle count register
 
-#if defined(ARDUINO_PORTENTA_H7_M7) || defined(ARDUINO_GIGA)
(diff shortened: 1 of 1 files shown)


chuckstanski:
dhalbert, I was able to get a hold of a Arduino GIGA R1 board.  There were some issues with running the code on the M4 core.  My latest code has been tested and works on the Arduino Nano BLE 33 as well as the Arduino GIGA R1 (both the M7 and M4 cores).

ladyada (approved):

ladyada:
thanks!

Issue #450, opened 2025-11-25
the ESP32-C2 ESP32-C61 /ESP8684 is not supported


EmileSpecialProducts:
I get the following error on the ESP32-C2, This is because the ESP32-C2 does not have a RMT.
Can the support be added ?

pioarduino And Arduino Release v3.3.4 

You can have a look at the [FastLed ](https://github.com/FastLED/FastLED)This does support the ESP32-C2 using the SPI to generate the data pulses.
See also [espressif/idf/led_strip](https://github.com/espressif/idf-extra-components/tree/master/led_strip)

```
.pio/libdeps/esp32-c2-devkitm-1-4MB/Adafruit NeoPixel/esp.c: In function 'espShow':
.pio/libdeps/esp32-c2-devkitm-1-4MB/Adafruit NeoPixel/esp.c:48:10: error: unknown type name 'rmt_data_t'
   48 |   static rmt_data_t *led_data = NULL;
      |          ^~~~~~~~~~
.pio/libdeps/esp32-c2-devkitm-1-4MB/Adafruit NeoPixel/esp.c:56:19: error: 'rmt_data_t' undeclared (first use in this function)
   56 |       led_data = (rmt_data_t *)malloc(requiredSize * sizeof(rmt_data_t));
      |                   ^~~~~~~~~~
.pio/libdeps/esp32-c2-devkitm-1-4MB/Adafruit NeoPixel/esp.c:56:19: note: each undeclared identifier is reported only once for each function it appears in      
.pio/libdeps/esp32-c2-devkitm-1-4MB/Adafruit NeoPixel/esp.c:56:31: error: expected expression before ')' token
   56 |       led_data = (rmt_data_t *)malloc(requiredSize * sizeof(rmt_data_t));
      |                               ^
.pio/libdeps/esp32-c2-devkitm-1-4MB/Adafruit NeoPixel/esp.c:69:9: error: implicit declaration of function 'rmtDeinit'; did you mean 'i2cDeinit'? [-Wimplicit-function-declaration]
   69 |         rmtDeinit(rmtPin);
      |         ^~~~~~~~~
      |         i2cDeinit
.pio/libdeps/esp32-c2-devkitm-1-4MB/Adafruit NeoPixel/esp.c:81:14: error: implicit declaration of function 'rmtInit' [-Wimplicit-function-declaration]
[...]

PR #451, opened 2025-12-07
Added support for the SPI interface for the ESP32C2 

EmileSpecialProducts:
Added support for the SPI interface for the ESP32-C2 but can also be used for the other devices.

The code is standard only for the ESP32-C2 / C61 (When this will be supported by Arduino platform) 
But can also be enabled for the other devices when the #define ADAFRUIT_SPI is added.
This is tested on the ESP32-C3/S3/C2.
On one channel and two channels.
This will resolve the issue #450

### DIFF
--- a/esp.c
+++ b/esp.c
@@ -16,10 +16,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 #if defined(ESP32)
-
 #include <Arduino.h>
+#endif 
+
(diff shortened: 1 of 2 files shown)


EmileSpecialProducts:
I will close this pull request and open a new one as this version has a 6 us pule on the spi_bus_initialize command, and this is used every show command. And makes it also slow.

PR #454, opened 2026-01-28
Support timing for CH32V00x running at 48MHz

maxint-rd:
The library already has support for the CH32 family of microprocessors (thanks @hathach) , but it didn't work for the lower clock speed of 48 MHz that is used by the popular CH32V003 and its newer CH32V00x siblings.

As mentioned in [this commit](https://github.com/adafruit/Adafruit_NeoPixel/commit/9dd0fcfb4022c63fe370099ee42127471eda7e09) the CH32 family was already supported for frequencies of 56 MHz and higher. This PR decreases the number of "nop;" codes used in `ch32Show()` within [Adafruit_NeoPixel.cpp](https://github.com/adafruit/Adafruit_NeoPixel/blob/master/Adafruit_NeoPixel.cpp), to specifically support 48 MHz. For higher frequencies the total number of nops remains the same.

The modified code was tested to work fine using the examples "strandtest" and "simple",  with both short (8) and long (256) strings of Neopixels. Testing was done
on these CH32 processors, all running at 48MHz:
- CH32V002J4M6 (SOP8) / CH32V002D4U6 (QFN12)
- CH32V003J4M6 (SOP8)
- CH32V003A4M6 (SOP16)
- CH32V006F8P6 (TSSOP20)
- CH32V006E8R6 (QSOP24) 
- CH32V006K8U6 (QFN32)

More precise testing and measurement of timing was done use the CH32V006E8R6 with different compilation options. Timing was measured with a cheap 24MHz logical analyzer. I didn't see significant differences for either -Os, -Os with LTO or -Og (for debugging with symbols). Bit-timing was fairly close to the 1250ns specified for 800kHz Neopixels, with average byte-timing only a little over 10us.

Thanks @BitByteBeetle for his contribution to test this PR with all these processors.

### DIFF
--- a/Adafruit_NeoPixel.cpp
+++ b/Adafruit_NeoPixel.cpp
@@ -272,11 +272,13 @@ static void ch32Show(GPIO_TypeDef* ch_port, uint32_t ch_pin, uint8_t* pixels, ui
       // High 800ns
       *set = ch_pin;
       __asm volatile ("nop; nop; nop; nop; nop; nop; nop; nop;"
-//        "nop; nop; nop; nop; nop; nop; nop; nop;"
-//        "nop; nop; nop; nop; nop; nop; nop; nop;"
-//        "nop; nop; nop; nop; nop; nop; nop; nop;"
         "nop; nop; nop; nop; nop; nop; nop; nop;"
         "nop;"
+#if CH32_F_CPU >= 56000000
+        "nop; nop; nop; nop; nop; nop; nop; nop;"
+        "nop; nop; nop; nop; nop; nop; nop; nop;"
+        "nop; nop; nop; nop; nop; nop; nop; nop;"
+#endif
 #if CH32_F_CPU >= 72000000
         "nop; nop; nop; nop; nop; nop; nop;"
         "nop; nop; nop; nop; nop; nop;"
@@ -300,8 +302,10 @@ static void ch32Show(GPIO_TypeDef* ch_port, uint32_t ch_pin, uint8_t* pixels, ui
       // Low 450ns
       *clr = ch_pin;
       __asm volatile ("nop; nop; nop; nop; nop; nop; nop; nop;"
-//        "nop; nop; nop; nop; nop; nop; nop; nop;"
         "nop;"
+#if CH32_F_CPU >= 56000000
+        "nop; nop; nop; nop; nop; nop; nop; nop;"
+#endif
 #if CH32_F_CPU >= 72000000
         "nop; nop; nop; nop; nop; nop; nop; nop; nop;"
 #endif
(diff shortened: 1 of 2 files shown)


hathach (approved):
perfect, thank you. This look good to me. Note I didn't test on actual hardware and/or measurement tool.

maxint-rd:
Hello @hathach ,

Excellent, thank you for merging this PR. 
**TLDR; Testing showed the new timing doesn't work for CH32V203 running @ 48 MHz. More testing/fixing is needed.**

For completeness sake I wanted to test some more frequencies on real hardware, and measure timings with the logical analyzer. For higher frequencies I only have the CH32V203 (TSSOP20) in my collection. It supports clock speeds of 48MHz up to 144MHz using the internal oscillator, i.e. the same frequencies that are supported by the library. Note that some CH32 processors also have lower internal clock speeds that are not supported (eg. V00x has 8 and 24MHz; X03x has 8, 12, 16 and 24MHz).

For this test I used an 8-pixel strip and the "simple" example with some added serial printing, compiled using -Os (default) optimization. The example worked at all tested frequencies of the V203, except for 48MHz, which is remarkable. However, even more remarkable is that the 56MHz timing for V203 running @ 48 MHz is working, even though it's a bit too slow.
Could it be that the nop execution on V203 is faster than on any of the tested V00x processors? Do NOPs have different duration for the various RV CPUs? Are there other differences between V00x and V203 resulting in this behavior? Probably more testing is needed...

For each frequency I measured the byte-duration. A full byte for an 800kHz NeoPixel should take 10us plus some extra time after the last bit. Measurements may be inaccurate due to the fairly low sampling rate of my analyzer (max. 24MHz). 
The table below shows timings from the start of the first byte up to the start of the next byte.

|clock speed|byte length|bit frequency|0 / 1 pulse length|
|--------|--------|--------|--------|
|144MHz|9.83-9.92us|774-828kHz|0.375us / 0.833us|
|120MHz|10.13-10.17us|774-800kHz|0.417us / 0.833us|
|96MHz|9.79-9.916us|774-857kHz|0.417us / 0,792us|
|72MHz|9.88-10.04us|774-828kHz|0.417us / 0,792us|
|56MHz|9.95-10.25us|800-828kHz|0.417us / 0,792us|
|48MHz <sup>1</sup>|5.16-5.83us|1200-1412kHz|0.17us / 0.50us|
|48MHz <sup>2</sup>|11.63-11.96us|667-685kHz|0.375us / 0.875us|

<sup>1</sup> 48MHz timing for V00x is not working for V203 @ 48MHz
<sup>2</sup> 56MHz timing is working for V203 running @ 48MHz

For now I'll call it a day, but I'd like to do a bit more testing, such as measuring running a bunch of NOPs on both V00x and V203. Perhaps also test the X033 which I have in my collection.
In the mean time I could submit an update to my PR that uses different 48MHz timing for V203.

BTW. I saw that the CH32fun project uses a somewhat different method that looks at the number of systicks passed. Don't know if that method would work better or not.

hathach:
perfect, thank you very much for your detail explanation. NOP indeed can take different time on different arch i.e IPC (instructions per cycle) e.g on m4 with instruction pipeline it can run faster with the same clock freq in comparison with M0. I am not familliar with riscv and v203 v00x but it may be the case here. At higher freq clock, the diff may be small enough to be withing timing interval. **I think the simple fix with ifelse @48mhz for cpu is good enough.** 

PS: using systick consume the same amount of blocking time with simpler code, however, systick maybe configured differently by user and may not available on all ports.

Issue #441, opened 2025-07-31
Interrupts are disabled for the RP2040 during show()


jmamma:
Given the rp2040 implementation is using PIO, do interrupts need to be disabled during show()?

Bodmer:
Corrected in pull request. See:

https://github.com/adafruit/Adafruit_NeoPixel/pull/455

PR #455, opened 2026-02-10
RP2040 interrupts do not need to be disabled as library uses PIO

Bodmer:
The interrupts do not need to be disabled during show() for the RP2040 as hardware PIO generates the timing signals.

Tested on Arduino Nano RP2040 Connect board with Arduino-pico board package.

Solves:
https://github.com/adafruit/Adafruit_NeoPixel/issues/441


BlinxFox:
Do you still need to enable interrupts afterwards?

